PR #2, opened 2014-08-07
User/nnz/port tom ice ridging

nikizadehgfdl:
This branch tries to "port" the development of ridging scheme from Torge Martin in SIS1 into SIS2.
The model compiles and runs (in debug mode too). It reproduces dev/master timestats.

To turn on the scheme do a  SIS_override with 
DO_RIDGING = True 

Without having a test case at hand I have no way to evaluate the scheme.
Even though it produces non-zero ridging rate, the timestat (after 1 day) is identical to the original (with  DO_RIDGING = False)

This pull request needs critical review and evaluation!


Hallberg-NOAA:
This merge is now too old to merge efficiently, and the code that was introduced has already been ported into SIS2, so this pull request is being closed.

PR #32, opened 2016-05-16
Ice tracer flow control and ice age examples

ashao:
The ice model now has a capability for tracers to be advected along with the ice both horizontally and between ice thickness categories. The tracer flow control structure is associated with the IST variable as 'SIS_tracer_flow_CSp'. Tracer stocks are not yet implemented, though some initial code is available. The bulk of this code was ported over from MOM_tracer_flow_control and so the overall behavior is similar between SIS and MOM6. 

Two ice-age tracers with slightly different boundary conditions have been implemented in ice_age_tracer.F90 that can be used as a template for other tracer packages. These were based on the ideal age tracer code from MOM6. The first tracer, referred to as an ice_age_areal simply increments the age of the ice by the length of the timestep if the ice mass in that category exceeds  a minimal amount. The second tracer ice_age_mass, has the same source of age, but also includes a sink of age when new ice is formed. See Hunke and Bitz [2009], JGR and Lietaer et al. [2011], Ocean Modeling for similar implementations in other models. These tracers were tested in the Baltic case for two years and for 10 years in the global ocean-ice coupled case. These tracers have not been validated for scientific use.


adcroft:
I like the name "ice-age tracer". :+1:

ashao:
A name so nice I committed it twice then thrice. (Certainly had nothing to do with the fact that I couldn't keep my own branches straight).

StephenGriffies:
cool indeed...

On Tue, May 17, 2016 at 6:24 AM, Alistair Adcroft <notifications@github.com>
wrote:

> I like the name "ice-age tracer". [image: :+1:]
>
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/NOAA-GFDL/SIS2/pull/32#issuecomment-219536567>
>



-- 
Dr. Stephen M. Griffies
NOAA Geophysical Fluid Dynamics Lab
201 Forrestal Road
Princeton, NJ 08542
USA

ashao:
For reference, here's the output of the areal ice-age from a SIS2 (1 degree, ocean-ice normalyear forcing) run.

image

Issue #42, opened 2016-08-29
SIS2 is causing answers mismatch across restarts in dev/master/2016.08.23


nikizadehgfdl:
With the release tag of 2016.08.23 ALL CM4 and OM4 experiments I tried failed to reproduce across a restart. These include both MOM6_GOLD_SIS2_nobergs and MOM6_GOLD_SIS2_bergs.

When I retract SIS2 to the commit corresponding to the previous release tag 2016.06.29  (eab89a62a4c1) the experiments reproduce across restart again.

So something in the rapid development of SIS2 since 6/29 is causing a restart issue.

nikizadehgfdl:
I tracked down this answers mismatch across restarts to commit ( abb82c39b9 ).

Hallberg-NOAA:
I had previously tested the restarts with the  coupled_AM2_LM3_SIS2/AM2_SIS2_MOM6i_1deg test case, and found that they worked. I have retested this with the latest commit and found that it still reproduces across restarts (a series of 10 1-day runs give identical answers to a 10-day run). At this point, I suspect that this issue is restricted to dynamic memory allocation but not static.

nikizadehgfdl:
To fix the restart issues in dev/master I had to 

1. Move "call set_domain(G%Domain%mpp_domain)" up before "call ice_data_type_register_restarts()"

2. Move "call nullify_domain()" down after "call icebergs_init()" (it could go sooner than that, I am going to try).

This fixed the restart issues for all the models I tried. 
I'll make a pull request shortly.

nikizadehgfdl:
Fixed.

PR #45, opened 2016-08-31
Fix restart issue #42, set_domain and nullify_domain

nikizadehgfdl:
- Fixes restart mismatch issue #42 
- Moving set_domain and nullify_domain calls


nikizadehgfdl:
I tried a few of the ocean-ice tests as well as the CM4 experiments and saw no answer changes from this update.

PR #47, opened 2016-09-23
1) Three new fields have been create: ustar_berg, area_berg and mass_berg

sternalon:
These three fields are created in the ice_data_type, and passed through to the ice_ocean_boundary_type, so that they can be passed through the coupler to the ocean. The goal is that these fields should affect the upper ocean boundary condition (in the same way that the ice shelf does). These new fields are only created and passed to the ocean when the flag PASS_ICEBERG_AREA_TO_OCEAN=.True.

2) The sea surface salinity has been added to the icebergs_run call, which allows the iceberg code to have access to the SSS.


Hallberg-NOAA:
After a discussion on our coding strategy, Alon has agreed that it would be a good idea to revise this to use optional arguments to the icebergs_run call to pass the iceberg mass and other fields back and for, so that the SIS2 code never has to access anything inside of the iceberg control structure.  The functionality of this pull request will be replaced by this newer code once it is ready.

PR #48, opened 2016-09-26
Enable calling the depostion fluxes after atmos_down

nikizadehgfdl:
- The depostion flux treatment needs to be called at a different time
  than the gas flux treatment because they are generally calulated
  as part of atmos_down and atmos_up processes which are not yet done
  at the time the gas flux calculation is called.
- This commit introduces a new ice interface that can be called anywhere
  in the fast atmos loop, particularly after update_ice_model_fast
  and it sums the deposition fluxes similar to the ice_model subroutine sum_top_quantites
- This update should be used along with the coupler updates in gitlab
  branch user/nnz/fix_esm_dust_flux to give consistent results for the
  (dry and wet) depostion fluxes for ESM.


nikizadehgfdl:
This pull request is for keeping my branch up-to-date with dev/master. 
Please do not handle it yet!  It needs a newer version of the coupler.

PR #54, opened 2016-11-08
Update fluxes from ice to ocean when ice is specified

ashao:
For coupled runs where the ice is specified, fluxes calculated during the fast thermodynamics step were not being passed to the ice ocean flux structure. The changes here update the ice-ocean fluxes during the slow thermodynamics step. Because these changes within the logical block when 'specified_ice' is true, this should not affect the behavior of coupled runs with dynamic ice. Answers did not change for the Baltic and Baltic_ALE_z cases.

### DIFF
--- a/SIS2_ice_thm.F90
+++ b/SIS2_ice_thm.F90
@@ -99,10 +99,6 @@
 
 module SIS2_ice_thm
 
-! for calling delta-Eddington shortwave from ice_optics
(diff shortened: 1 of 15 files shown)


ashao:
After discussion with @Hallberg-NOAA, some lines from SIS2_thermodynamics were moved up one level to slow_thermodynamics to avoid duplicating code. The logic in slow_thermodynamics was updated accordingly. This change does not change answers in the Baltic and Baltic_ALE_z test cases.

PR #88, opened 2018-07-20
Added stress_mag to ice_data_type

Hallberg-NOAA:
Added a new element, stress_mag, with the time-mean of the magnitude of the
wind stresses at tracer points, to the publicly visible ice_data_type.  It is
not yet being allocated or used, so all answers are bitwise identical.


Hallberg-NOAA:
Please merge in  this PR quickly so these minor interface changes make the Xanadu release.  Thank you.

PR #97, opened 2018-12-13
Fix ice restore issues

nikizadehgfdl:
Fix the ice restoring issues
    
    - This update addresses two independent issues with the ice restoring mechanism:

      1. SM4p model blows up with too much ice when it tries to do ice restoring to climatology.
         This only happens when SIS1 is replaced by SIS2.
         Upon investigation I found that the algorithm in SIS_slow_thermo.F90 has an update ordering issue.
         At the point when qflx_res_ice is calculated it updates the heat input to ice through bheat.
         But this does not to have an immediate direct effect on the ice thickness
         and its effect is delayed till the next timestep through other fluxes (flux_sh_ocn_top).
         This leads to a runaway increasing of the ice thickness (wrong feedback).
         To get around this we can increase bmelt instead of bheat which will decrease the ice thickness
         subsequently in the same timestep , in the same module through call ice_resize_SIS2().
         Standalone tests shows this has the correct restoring effect on the ice.
    
      2. SM4p has a restart issue, the answers of straight runs and restarted runs do not match (1x8days != 2x4days)
         I tracked this to a bit of code in ice_model.F90 that applies land mask to mH_ice.
         We subsequently tried applying lad mask to part_size (PR !95). It fixed a NaN issue but did not
         fix the restart issue.
         Applying masks to state variables such as mH_ice is leading to the restart issue because these
         state variables are being modified only after restarts and not in straight (no-restart) runs.
         I provided a switch to avoid this masking if the user needs to. With it SM4p reproduces.

### DIFF
--- a/src/SIS_slow_thermo.F90
+++ b/src/SIS_slow_thermo.F90
@@ -742,9 +742,24 @@ subroutine SIS2_thermodynamics(IST, dt_slow, CS, OSS, FIA, IOF, G, IG)
       qflx_res_ice(i,j) = -(LatHtFus*Rho_ice*Obs_h_ice(i,j)*Obs_cn_ice(i,j,2)-e2m_tot) / &
                            (86400.0*CS%ice_restore_timescale)
       if (qflx_res_ice(i,j) < 0.0) then
+        !There is less ice in model than Obs,
+        !so make some ice by increasing frazil heat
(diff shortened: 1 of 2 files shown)


Hallberg-NOAA:
This PR introduces a new runtime parameter, APPLY_MASKS_AFTER_RESTART, which causes the SIS_parameter_doc.all files to change.  By default answers are bitwise identical in the MOM6_examples test cases.

PR #105, opened 2019-07-16
Bug fix for do_ice_limit

nikizadehgfdl:
salt_to_ice already includes the IST%part_size factor at line 1197.

This extra factor (with an effectively out of bound k=ncat+1) causes the ice limiting simulations (sm4p1) to crash for some processor layouts with:
FATAL from PE   812: NaN in input field of reproducing_sum(_2d).
where the reproducing_sum refers to the ice salinity.

I verified that this mod fixes the issue with sm4p1 and luckily reproduces the answers for the layouts that did work before the fix.


adcroft:
The fix looks right to me. @Hallberg-NOAA do you have a config that can actually test this?

adcroft:
@Hallberg-NOAA Bump

Issue #107, opened 2019-08-13
SIS2 in SPECIFIED_ICE mode (used in AM4) does not reproduce across restarts


nikizadehgfdl:
There is a push to replace SIS1 with  SIS2 in AM4 workflow. The model runs, however it does not reproduce across restarts. 

```
xml: /ncrc/home2/Niki.Zadeh/xmls/xanadu/mdt_xml_dev/ESM4_amip_sis2_D1.xml
exp: ESM4_amip_sis2_D1
```
FYI, it doesn't reproduce the AM4 answers with SIS1 either. After fixing this bug, we will need to make a long run to evaluate these new answers with SIS2.

nikizadehgfdl:
This is fixed in dev/gfdl.

Issue #101, opened 2019-04-01
Nice if SIS2 indicated that it read restarts


nikizadehgfdl:
SIS2 does not  print out in stdout that it was initialized from restart files. 
MOM6 does:
```
NOTE from PE     0: MOM_restart: MOM run restarted using : INPUT/MOM.res.nc
NOTE from PE     0: MOM_restart: MOM run restarted using : INPUT/MOM.res_1.nc
NOTE from PE     0: MOM_restart: MOM run restarted using : INPUT/MOM.res_2.nc
```
This confuses users (including yours truly).
It would be nice if SIS2 indicated that it read restarts.

Assign this to me and I'll take care of it.

nikizadehgfdl:
This is implemented. Thanks. 
```
NOTE from PE     0: callTree:       ---> ice_model_init():restore_from_restart_files ice_model.res.nc
```

PR #109, opened 2019-08-21
Fix for specified ice restart issue #107

nikizadehgfdl:
- Variable Ice%sCS%OSS%SST_C is not an ice state and hence is not in ice_model.res.nc
  However for SPECIFIED_ICE=True it should be updated after restart, otherwise the model
  would have a restart issue #107 
  In specified ice this variable is read from data_table every time step. So that is also
  the easiest way to update it after a restart.
- This update resolves issue #107 .

- This update also addresses issue #101 by adding some indication 
  that SIS2 is being restored from restart files.
  Currently with input_filename being 'n' or 'r'  having no effect, there is no
  indication in stdout whether restart files are being read or not.
  This would make it a little easier to debug restart issues.

### DIFF
--- a/src/ice_model.F90
+++ b/src/ice_model.F90
@@ -2500,11 +2500,10 @@ subroutine ice_model_init(Ice, Time_Init, Time, Time_step_fast, Time_step_slow, 
       !However it is not an ice state variable and hence is not in ice restarts.
       !Updating it from specified_ice data avoids restart issues for such models.
       if (specified_ice) then
-        allocate(dummy2d(sG%isc:sG%iec,sG%jsc:sG%jec))
+        allocate(dummy2d(isc:iec,jsc:jec))
         allocate(dummy3d(isc:iec,jsc:jec,2))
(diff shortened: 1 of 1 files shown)


Hallberg-NOAA:
The added code uses a mixture of isc and sG%isc, but I am concerned that isc (and related variables) are not defined here.  Was this use of different variables deliberate, @nikizadehgfdl, and if not could these variables be treated the same?

nikizadehgfdl:
I followed the syntax of the existing call on [line 2527](https://github.com/NOAA-GFDL/SIS2/blob/9991d35038fc445b6a79d56fa2bf5ecd47c5571b/src/ice_model.F90#L2527) where the 2nd and 3rd arguments are on isc,jsc whereas the 4th argument is allocated on sC%isc.
I assumed that isc,jsc where defined since they are being used on previous lines, e.g. [line 2441](https://github.com/NOAA-GFDL/SIS2/blob/9991d35038fc445b6a79d56fa2bf5ecd47c5571b/src/ice_model.F90#L2441).

nikizadehgfdl:
I fixed the dummy arrays extents without changing answers.

PR #112, opened 2020-03-09
slow/fast mask bug fix

raphaeldussin:
t_skin is masked with slow grid mask instead of fast grid. Results in seg fault.

@Hallberg-NOAA @adcroft @marshallward


Hallberg-NOAA:
I have examined this change and concur that it is correct.

PR #115, opened 2020-03-24
SIS2: +Remove carriage returns in parameter descriptions

Hallberg-NOAA:
Removed hard-coded newline symbols from most SIS2 parameter descriptions in
calls to get_param or log_param, instead allowing these routines to format the
output automatically.  All answers are bitwise identical, but there are changes
in many of the variable descriptions in the SIS_parameter_doc files.


marshallward:
This passed my regression test.

(But alas, I still cannot merge)

Hallberg-NOAA:
I am merging this PR, despite the fact that I introduced it, because @marshallward has reviewed and approved it, and would have merged it had the permissions been set appropriately.

Note that when MOM6-examples is updated to use this PR, the SIS_parameter_doc files there will also need to be updated.

PR #116, opened 2020-04-23
+Corrected SIS2 documentation problems

Hallberg-NOAA:
Corrected documented units in 6 diagnostics and 5 input parameters, removed an
unused parameter, and moved 3 other parameters so that they first appears in the
module where they are actually used.  Also fixed the values sent to the
diagnostic MIS_tot to include the snow mass.  All answers are bitwise identical,
but there are changes to the SIS_available_diags and SIS_parameter doc files.


marshallward:
I have run the Gaea regression test locally and it passed with no errors.

PR #117, opened 2020-05-21
User/jml/add new fms io 02

wrongkindofdoctor:
I'm submitting this PR so that the developers so that developers can easily see the changes I've made to the restart interfaces and (hopefully), assist with issues I'm having reproducing answers from restart runs.


Hallberg-NOAA:
The new PR #144 extensively revises the SIS2 restart capability to work via the code in the MOM6 src/framework and config_src/infra directories. This new code reproduces answers across restarts, either for single-file or decomposed restart files, and does so when the calls eventually resolve to either the FMS2 or the original FMS routines, depending on which directory under config_src/infra/FMS[12] is used. This draft is therefore no longer relevant to SIS2, although it was helpful for identifying the changes in the I/O interfaces between FMS2_io and the original fms interfaces.

PR #119, opened 2020-05-24
SIS2: +(*)Add sea ice initialization options

Hallberg-NOAA:
This group of commits adds several new options for initializing the state of
the sea ice, refactors related parts of the code for simplicity or cleanliness,
and makes a public interface more consistent with other SIS2 code.  There are
some atypical configurations in which answers could change at roundoff due to
changing order of arithmetic.  All answers in the MOM6-examples test cases are
bitwise identical, and there are new entries in the SIS_parameter_doc files. The
commits in this PR include:

- NOAA-GFDL/SIS2@e00bf47 +Use SIS_state_initialization for initialization
- NOAA-GFDL/SIS2@8fd108f +Added SIS_state_initialization.F90
- NOAA-GFDL/SIS2@7c04797 Initialized IST%t_surf to 0 deg C in Kelvin
- NOAA-GFDL/SIS2@d1146ba +Rescale thickness in rescale_ice_state_rest_files
- NOAA-GFDL/SIS2@a118edd (*)Refactored ice_model_init
- NOAA-GFDL/SIS2@b53a904 +(*)Changed interface to get_sea_surface
- NOAA-GFDL/SIS2@2d6cc20 (+)Added ridging diagnostic when do_ridging is true

### DIFF
--- a/src/SIS_ctrl_types.F90
+++ b/src/SIS_ctrl_types.F90
@@ -253,9 +253,6 @@ subroutine ice_diagnostics_init(IOF, OSS, FIA, G, US, IG, diag, Time, Cgrid)
   if (coupler_type_initialized(IOF%tr_flux_ocn_top)) &
     call coupler_type_set_diags(IOF%tr_flux_ocn_top, 'ice_model', diag%axesT1%handles, Time)
 
-!### THIS DIAGNOSTIC IS MISSING.
-!  XYZ%id_strna    = register_SIS_diag_field('ice_model', 'STRAIN_ANGLE', diag%axesT1,Time, &
-!               'strain angle', 'none', missing_value=missing)
 
   FIA%id_sw_dn   = register_SIS_diag_field('ice_model', 'SWDN', diag%axesT1, Time, &
                'Downward shortwave heat flux at the bottom of the atmosphere', &
@@ -344,7 +341,9 @@ subroutine ice_diagnostics_init(IOF, OSS, FIA, G, US, IG, diag, Time, Cgrid)
   if (coupler_type_initialized(OSS%tr_fields)) &
     call coupler_type_set_diags(OSS%tr_fields, 'ice_model', diag%axesT1%handles, Time)
 
-!### THIS DIAGNOSTIC IS MISSING.
+!  These are omitted diagnotics with no imminent plans to add them.
+!  XYZ%id_strna = register_SIS_diag_field('ice_model', 'STRAIN_ANGLE', diag%axesT1,Time, &
+!               'strain angle', 'none', missing_value=missing)
 !  XYZ%id_obi   = register_SIS_diag_field('ice_model', 'OBI', diag%axesT1, Time, &
 !       'ice observed', '0 or 1', missing_value=missing)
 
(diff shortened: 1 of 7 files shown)


kshedstrom:
I approve this PR - heck, I asked for it!

kshedstrom:
I should add that although this appears to work when providing an ice thickness and concentration, an hour into the run I have very little ice and it is growing hour by hour, as if I started with none. I could see non-zero ice during initialization in the debugger.

kshedstrom:
Never mind, wrong units in my file. Just off by a factor of 900.

Hallberg-NOAA:
This PR currently offers a rescaling factor for salinity (e.g., so that ocean salinity can be easily used to roughly convert to sea ice salinity by rescaling by a factor of order 1/7). @kshedstrom, would you like to see runtime rescaling factors for the other variables that are read from files? (Although if we did this, I would not suggest setting the default to be a factor of 900!)  Probably the best approach if this is desired would be by following this PR up with another PR.

kshedstrom:
You should probably change ARCTIC to ANTARCTIC here:

>   call get_param(PF, mdl, "ARCTIC_ICE_EDGE_IC", Antarctic_ice_edge, &
                 "The northern latitude of Antarctic ice in an initial condition.", &
                 default=-91.0, units="degrees of latitude", do_not_log=just_read)

I'm not out of the woods on the initialization attempts, but will explore more next.

kshedstrom:
OK, the initialization works. Then in SIS2_thermodynamics in a loop ending at SIS_slow_thermo.F90L#1016, all the ice is eaten.

Hallberg-NOAA:
How did you initialize the sea-ice temperature?

  As currently written, the default is to use a uniform temperature of 0 C, which is above freezing so it all melts.  I could change the default to -4 C or something, at which point the ice would stick around and might make more physical sense, but this would violation one of @adcroft's guidelines discouraging the use of arbitrary dimensional default values. 0 C is not very physical, but at least it is not arbitrary!

kshedstrom:
Thanks, yes, but you need this too:
```
diff --git a/src/SIS_state_initialization.F90 b/src/SIS_state_initialization.F90
index 2217716..264d4b1 100644
--- a/src/SIS_state_initialization.F90
+++ b/src/SIS_state_initialization.F90
@@ -284,7 +284,7 @@ subroutine ice_state_thermo_init(IST, Ice, G, IG, US, PF, init_Time, just_read_p
             " \t\t specified depression below the bulk ice freezing point \n"//&
             " \t file - Read sea ice temperatures or enthalpies from a specified file \n"//&
             " \t data_override - use the data_override capability or freezing enthalpy everywhere.", &
-            default='uniform', do_not_log=just_read)
+            default='uniform_temp', do_not_log=just_read)
 
   call get_param(PF, mdl, "SNOW_ENTHALPY_INIT_CONFIG", enth_snow_config, &
             "A string that determines how the snow enthalpy is initialized for a new run: \n"//&
@@ -293,7 +293,7 @@ subroutine ice_state_thermo_init(IST, Ice, G, IG, US, PF, init_Time, just_read_p
             " \t\t specified depression below the bulk ice freezing point \n"//&
             " \t file - Read sea ice temperatures or enthalpies from a specified file \n"//&
             " \t data_override - use the data_override capability or freezing enthalpy everywhere.", &
-            default='uniform', do_not_log=just_read)
+            default='uniform_temp', do_not_log=just_read)
 
   call get_param(PF, mdl, "ICE_BULK_SALINITY", ice_bulk_salin, &
                  "The fixed bulk salinity of sea ice.", units = "g/kg", &
```
or else it doesn't even look for temperature. As for 0 C, it seems arbitrary from a Kelvin perspective...

Then Alistair's makefile doesn't rebuild on changes to SIS2.

Hallberg-NOAA:
Thanks for the suggested corrections, @kshedstrom, the recent update to this PR should have implemented them.

PR #123, opened 2020-06-12
(*)Updated SIS2 parameter defaults

Hallberg-NOAA:
Updated the default values for CGRID_ICE_DYNAMICS (to "C") and
INCONSISTENT_COVER_BUGS and CFL_MASS_NEGLECT_BUG to false, and made the
documented units of MIN_POND_FRAC and MAX_POND_FRAC consistent across the SIS2
code.  In addition, component arguments were added to the close_param_file calls
so that the parameter file defaults are properly documented.  By design, all
solutions in the MOM6-examples test suite are bitwise identical, but answers can
change for other cases, and there are changes in the SIS_parameter_doc files.

### DIFF
--- a/src/SIS_dyn_trans.F90
+++ b/src/SIS_dyn_trans.F90
@@ -2122,7 +2122,7 @@ subroutine SIS_dyn_trans_register_restarts(mpp_domain, HI, IG, param_file, CS, &
   endif
   allocate(CS)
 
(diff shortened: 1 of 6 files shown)


Hallberg-NOAA:
Please note that https://github.com/NOAA-GFDL/MOM6-examples/pull/318 needs to be handled before this PR.

PR #124, opened 2020-06-18
+Improve SIS2 checksum capabilities for debugging

Hallberg-NOAA:
Improved the SIS2 checksum capabilities that are triggered when DEBUG=True,
especially when debugging restarts.  Some variables that should not necessarily
reproduce are not longer being reported, while other types that do need to
reproduce are being checked.  These changes include the new subroutines
OSS_chksum and TSF_chksum, along with new optional arguments to IOF_chksum and
Ice_public_type_chksum.  Also, the albedos and absorbed radiation diagnostics
are now being set to 0 in categories with 0 part size, whereas they had
previously just retaine their previous value; this does not change answers but
it does make debugging easier.  All answers and output files are unchanged.


marshallward:
This has passed my Gaea regression with no changes.

PR #125, opened 2020-06-22
This is the minimum update required to use 2020.01 FMS and coupler

nikizadehgfdl:
- Add domain position arguments to diag_axis_init calls
diag_axis_init has a new optional argument to specify the corner location of the diagnostics on the grid.
Although these are optional args the model would crash if it sends such corner diagnostics without passing the appropriate optional argument to diag_axis_init.


nikizadehgfdl:
This updates requires FMS 2019.01.02 tag or later.

Hallberg-NOAA:
This PR passes the relevant regression tests with bitwise identical answers, but it requires a version of FMS with the 2019.01.02 or 2020.01 tags or later.

PR #128, opened 2020-08-02
Combine all global sums in SIS_sum_output

Hallberg-NOAA:
Revised SIS_sum_output.F90 to merge all blocking global sums in SIS2 that are
used routinely into a single blocking sum in SIS_sum_output for improved model
efficiency.  Also removed unneeded variables from the SIS_sum_out_CS and
reformatted the module use statements in this module.  All answers are bitwise
identical.


marshallward:
This has passed my internal Gaea regression.

PR #129, opened 2020-08-03
Use MOM6 cpu_clock timers in place of mpp_clock

Hallberg-NOAA:
Replace calls to mpp_clock routines from mpp_mod with calls to cpu_clock from
the module MOM_cpu_clock, to limit the direct use of fms routines and expand the
standardized use of the MOM6 infrastructure.  Also corrected the granularity of
two of the sea-ice clocks, to avoid double counting with clocks at the same
level of granularity, and added clocks around calls that to dierctly exchange
information between the ice and ocean.  All answers are bitwise identical.


marshallward:
Internal regression testing has passed.

PR #130, opened 2020-08-05
+Simplified the logic about setting ocean albedos

Hallberg-NOAA:
Cleaned up the logic for when to reset the ocean albedo, and split the single
internal interface for set_ocean_albedo into set_ocean_albedo_from_astronomy and
set_ocean_albedo_from_coszen.  Also removed the element frequent_albedo_update
from the ice_rad_type, as it is no longer used.  Also cleaned up the module use
statements at the top of ice_model.F90 and SIS_types.F90, alphabetizing entries
and removing references to unused routines and types.  All answers are bitwise
identical.

### DIFF
--- a/src/SIS_types.F90
+++ b/src/SIS_types.F90
@@ -3,22 +3,18 @@
 !! these types, including allocation, deallocation, registration for restarts, and checksums.
 module SIS_types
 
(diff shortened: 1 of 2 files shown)


marshallward:
This has passed internal regression testing on Gaea.

PR #134, opened 2020-09-29
Fix data_override crash in interspersed ice mode

nikizadehgfdl:
CM4 highres in Interspersed Ice mode crashes when there is no ice_model.res.nc 
The crash happens on the two  calls to data_override_unset_domains on lines 155 and 315 because
  data_override is not initialized yet **on the fast ICE cores** that these ice calls run on.

I verified that this update fixes the issue.


Hallberg-NOAA:
Actually, these added calls are only executed on slow_ice_PEs, but I agree that they do address the issues with the data_override not being initialized.

PR #136, opened 2021-01-21
Lemieux landfast ice

kshedstrom:
This change compiles and runs, but the results look a little odd. I'd love for @Hallberg-NOAA to look it over.

The change to ice_model.F90 is what Liz needed for restarts. I shouldn't get the credit for it, but I haven't seen it go in yet.

### DIFF
--- a/src/SIS_dyn_cgrid.F90
+++ b/src/SIS_dyn_cgrid.F90
@@ -632,7 +632,7 @@ subroutine SIS_C_dynamics(ci, mis, mice, ui, vi, uo, vo, fxat, fyat, &
   real :: m_neglect2 ! A tiny mass per unit area squared [R2 Z2 ~> kg2 m-4].
   real :: m_neglect4 ! A tiny mass per unit area to the 4th power [R4 Z4 ~> kg4 m-8].
   real :: sum_area   ! The sum of ocean areas around a vorticity point [L2 ~> m2].
-  real :: drag_bot_u, drag_bot_v ! The speed at u and v points, respectively [L T-1 ~> m s-1].
+  real :: drag_bot_u, drag_bot_v ! A drag with the bottom at u & v points [R Z T-1 ~> kg m-2 s-1].
 
   type(time_type) :: &
     time_it_start, &  ! The starting time of the iteratve steps.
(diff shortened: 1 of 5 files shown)


src/SIS_dyn_cgrid.F90
```diff
@@ -996,6 +1029,8 @@ subroutine SIS_C_dynamics(ci, mis, mice, ui, vi, uo, vo, &
       v2_at_u =  CS%drag_bg_vel2 + 0.25 * &
                      (((vi(i,J)-vo(i,J))**2 + (vi(i+1,J-1)-vo(i+1,J-1))**2) + &
                       ((vi(i+1,J)-vo(i+1,J))**2 + (vi(i,J-1)-vo(i,J-1))**2))
+      if (CS%lemieux_landfast) v2_at_u_min = &
+                     min(vi(i,J), vi(i+1,J-1), vi(i+1,J), vi(i,J-1))**2
```
Hallberg-NOAA, line 1033: If these are all positive, this picks out the minimum magnitude, but if these are all negative this picks out the largest magnitude.  Maybe there should be absolute values to these velocities, or they should be squared before taking the minimum.

src/SIS_dyn_cgrid.F90
```diff
@@ -1024,13 +1059,22 @@ subroutine SIS_C_dynamics(ci, mis, mice, ui, vi, uo, vo, &
         drag_u = cdRho * sqrt(uio_init**2 + v2_at_u )
       endif
       if (drag_max>0.) drag_u = min( drag_u, drag_max )
+      if (CS%lemieux_landfast) drag_bot_u = CS%Tb_u(I,j) / &
```
Hallberg-NOAA, line 1062: If drag_bot_u were simply combined with drag_u, there would not be a need for a duplicated block at line 1067.

src/SIS_dyn_trans.F90
```diff
@@ -453,6 +454,13 @@ subroutine SIS_dynamics_trans(IST, OSS, FIA, IOF, dt_slow, CS, icebergs_CS, G, U
           call set_wind_stresses_C(FIA, ice_cover, ice_free, WindStr_x_Cu, WindStr_y_Cv, &
                                    WindStr_x_ocn_Cu, WindStr_y_ocn_Cv, G, US, CS%complete_ice_cover)
 
+          if (CS%lemieux_landfast) then
```
Hallberg-NOAA, line 457: I don't understand the point of putting in the option to use a older, deprecated style of the code with a new option.  I think that that the Warsaw_sum_order branch should be eliminated here.

kshedstrom:
Thanks for pointing out the error in the min. It's been that way forever in ROMS so I don't recall the original intent. I switched it to an average for now.

If I combine drag_bot_u with drag_u it gets combined in the diagnostics too. That might be fine, I don't know.

If you want to get rid of Warsaw_sum_order, that's up to you. I don't know what's coming or going.

Hallberg-NOAA (approved):
I have inspected these changes and agree that they appear to be correct, and I have also run the revised code through our test suite and confirmed that the answers are unchanged (although as expected there are new entries in the SIS_parameter_doc files with C-grid dynamics).

PR #139, opened 2021-03-13
Only write checksums from root PE

Hallberg-NOAA:
Modified the routines that write out the checksums of various types related to
the sea ice so that only the root PE writes the checksums.  The modified
routines include ocn_ice_bnd_type_chksum, atm_ice_bnd_type_chksum,
lnd_ice_bnd_type_chksum, and ice_data_type_chksum.  In some large-PE tests cases
this leads to a reduction in the volume of output to stdout by over 90% without
any loss of information.  All solutions are bitwise identical, but there may be
minor white-space reformatting of the output in some cases.


marshallward:
Passed my independent testing.

PR #140, opened 2021-04-10
SIS2: Revise the SIS2 use of framework code

Hallberg-NOAA:
This PR contains a series of commits that coordinate the SIS2 infrastructure
calls with the MOM6 infrastructure calls and make calls to the underlying FMS
infrastructure via that MOM6 framework or config_src/infra modules.  All answers
and output are bitwise identical, but there are some interface changes, and the
version of MOM6 that is used must include the changes in MOM6 PR #1363 to
dev/gfdl (merged in on April 6, 2021).  The changes in this PR include:

- NOAA-GFDL/SIS2@b10799c Use EAST_FACE instead of EAST to denote positions
- NOAA-GFDL/SIS2@27d462a Eliminate unused module use entries
- NOAA-GFDL/SIS2@3e342e9 Use MOM6/config_src/infra in SIS_framework.F90
- NOAA-GFDL/SIS2@6b24054 Call data_override via MOM_data_override
- NOAA-GFDL/SIS2@783f937 +Use same_domain from MOM_domain_infra
- NOAA-GFDL/SIS2@d451bc6 SIS_diag_mediator goes via MOM_diag_manager_infra

Issue #96, opened 2018-12-05
SIS2 ignores the input_filename in input_nml


nikizadehgfdl:
@Hallberg-NOAA , @wfcooke , @MJHarrison-GFDL 
When I set input_filename = 'r'  AND do not provide any INPUT/ice_model.res.nc
the model does not complain (unlike MOM6 behavior).
Also when I set input_filename = 'n' AND provide a INPUT/ice_model.res.nc
the model does not complain and reads the restart (unlike MOM6 behavior).
This is a change in behavior that confuses users and may cause models to do something wrong for those who depend on that logic.

Please verify if this is the new feature or a bug.

Also, it would be helpful if the stdout or some log file showed whether ice restart file(s) were actually read in a run.

Hallberg-NOAA:
This issue was addressed by PR #141.

PR #141, opened 2021-04-10
(*)+Stop ignoring SIS2 namelist entries

Hallberg-NOAA:
Revised the handling of the SIS2 namelist entries so that they are no longer
ignored, addressing SIS2 issue #96.  This includes modification so that the
behavior when input_filename = 'F' is exactly the same as if it is 'n' if there
are no restart files in the restart_input_dir, or as if it is 'r' if the restart
files are there.  Previously there had been differences in the output files
generated, but the solutions were always if input_filename = 'F'.  The entries
in dirs%restart_output_dir are now used to guide where SIS2 writes its restarts,
whereas this had previously been ignored with the output always going into
"./RESTART/".  As a part of this change, a 5 routines have been imported from
MOM_restart.F90 to help determine when this is a new run in 'F' mode, and there
have been a number of new arguments added to the restart-related interfaces,
both to support these changes, and to align these interfaces with those of
MOM_restart.F90.  Answers can change unless input_filename = 'F', and there can
be changes in which output files are written and the presence of one entry in
the SIS_parameter_doc.all files.

  Despite appearances to the contrary, this is actually a relatively simple single-commit
PR (commit 5e5525c...) with just three files modified, that builds on SIS2 PR #140.

### DIFF
--- a/src/SIS_ctrl_types.F90
+++ b/src/SIS_ctrl_types.F90
@@ -12,8 +12,7 @@ use SIS_diag_mediator, only : SIS_diag_ctrl, post_data=>post_SIS_data
 use SIS_diag_mediator, only : register_SIS_diag_field, register_static_field
 use SIS_dyn_trans,     only : dyn_trans_CS
 use SIS_fast_thermo,   only : fast_thermo_CS
-use SIS_framework,     only : domain2D, CORNER, EAST, NORTH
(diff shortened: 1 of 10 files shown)


marshallward:
This passed Gaea regression after updating `input_filename` to `F` in the coupled models.

PR #142, opened 2021-04-15
+Created SIS_restart module

Hallberg-NOAA:
Moved the SIS2 restart-related types and interfaces into the new module
SIS_restart.  The MOM_domain type arguments to only_read_from_restarts were also
made mandatory, reflecting what will need to be done when the SIS_restart is
modified to use new interfaces that work via the MOM6 infrastructure routines.
All answers are bitwise identical, but there are minor argument changes and a
number of module use statements needed to be modified.


marshallward:
Latest update has a format string error:
```
/lustre/f2/dev/gfdl/Marshall.Ward/CI/Gaea-stats-MOM6-examples/MOM6-examples/src/SIS2/src/SIS_restart.F90:280:21:

   write(count_msg, '(I)') nf
                     1
Error: Nonnegative width required in format string at (1)
``` 
`(I0)` should work here.

marshallward:
I fixed the format string error.

This passed my regression test, but as we now know my test is inadequate for catching restart errors.  I will do some hand-tests and try to further confirm this test but AFAIK we won't really know until we run it through Gaea regression.

marshallward:
My restart tests now reproduce the original result.

PR #144, opened 2021-04-23
SIS2: Revise SIS2 restart capability

Hallberg-NOAA:
This PR contains a small handful of commits that revise the SIS2 restart
capability to read and write without using any FMS routines directly, which
enables the migration to use the new FMS2 I/O interfaces while still maintaining
the ability to run with the traditional FMS I/O interfaces.  All solutions are
bitwise identical when used with either of the MOM6 config_src/infra directories
(although there are some changes in the output file metadata), and the SIS2 runs
in MOM6-examples with the new restart files reproduce answers when either I/O
via the root PE or parallel I/O are used.

  This SIS2 PR requires the use of MOM6 code with the changes in https://github.com/NOAA-GFDL/MOM6/pull/1373 .

  The commits in this PR include:

- NOAA-GFDL/SIS2@0b80b63 Corrected call to open_file in open_restart_units
- NOAA-GFDL/SIS2@71edd8c Merge branch 'dev/gfdl' into FMS2_restarts
- NOAA-GFDL/SIS2@621219b +Revised SIS_restart to avoid direct calls to FMS

### DIFF
--- a/src/SIS_restart.F90
+++ b/src/SIS_restart.F90
@@ -3,42 +3,107 @@ module SIS_restart
 
 ! This file is part of SIS2. See LICENSE.md for the license.
 
-use coupler_types_mod, only : coupler_type_register_restarts
-use fms_io_mod,        only : restart_file_type, FMS1_register_restart=>register_restart_field
-use fms_io_mod,        only : save_restart_FMS1=>save_restart, FMS1_restore_state=>restore_state
-use fms_io_mod,        only : FMS1_query_initialized=>query_initialized
-
+use MOM_checksums,     only : chksum => rotated_field_chksum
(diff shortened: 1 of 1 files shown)


marshallward:
This has passed my local (non-restart) regression as well as a proxy restart test thrown together, based on OM4_025:

* One 12h run
* Two 6h runs using `input_filename` as `n` and `r`
* Two 6h runs using `input_filename` as `F`.

This is probably not sufficient for restart testing but should eventually be revealed in the Gaea regression testing.

marshallward:
Gaea regression test (post-merge): https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/12546

Issue #143, opened 2021-04-22
Indexing error if NK_ICE = 1


kshedstrom:
In src/SIS2_ice_thm.F90 at L369, there is an index out of bounds with temp_est(2) if there is only one vertical level. Perhaps there should be a check saying there has to be at least two levels.

Hallberg-NOAA:
This issue was addressed with PR https://github.com/NOAA-GFDL/SIS2/pull/145.

PR #145, opened 2021-04-26
Add branch to handle 1-layer ice thermodynamics

Hallberg-NOAA:
Added a logical branch in ice_temp_SIS2 to properly handle the case where
there is a single thermodynamic layer in the sea ice, addressing SIS2 issue
#143.  All answers are bitwise identical.


marshallward:
Passing local regression.

PR #148, opened 2021-05-19
FMS2 requires units for all netcdf dimension

nikizadehgfdl:
- All ocean-ice tests were failing to read from their own restart files because FMS2
  could not find "units" for a few new netcdf dims cat,cat0,z_ice,z_snow

- Mode of failure was
```
FATAL from PE    32: NetCDF: Attribute not found
Image              PC                Routine            Line        Source
fms_MOM6_SIS2_com  0000000002519CE1  Unknown               Unknown  Unknown
fms_MOM6_SIS2_com  000000000173F7FC  mpp_mod_mp_mpp_er          68  mpp_util_mpi.inc
fms_MOM6_SIS2_com  0000000001812E00  fms_io_utils_mod_         194  fms_io_utils.F90
fms_MOM6_SIS2_com  0000000001760277  netcdf_io_mod_mp_         337  netcdf_io.F90
fms_MOM6_SIS2_com  00000000007A7D5E  mom_io_infra_mp_g         502  MOM_io_infra.F90
fms_MOM6_SIS2_com  0000000000668784  sis_restart_mp_re        1240  SIS_restart.F90
fms_MOM6_SIS2_com  00000000004B0152  ice_model_mod_mp_        2311  ice_model.F90
fms_MOM6_SIS2_com  000000000040C5CB  coupler_main_IP_c        1697  coupler_main.F90
fms_MOM6_SIS2_com  0000000000401B42  MAIN__                    568  coupler_main.F90
```

### DIFF
--- a/src/SIS_restart.F90
+++ b/src/SIS_restart.F90
@@ -1047,14 +1047,15 @@ subroutine save_restart(directory, time, G, CS, IG, time_stamp)
   if (PRESENT(time_stamp)) restartname = trim(CS%restartfile)//trim(time_stamp)
 
   if (present(IG)) then
-    call set_axis_info(extra_axes(1), "cat", longname="Ice thickness categories", ax_size=IG%CatIce)
-    call set_axis_info(extra_axes(2), "cat0", longname="Ice thickness categories with open water", ax_size=IG%CatIce+1)
-    call set_axis_info(extra_axes(3), "z_ice", longname="Ice vertical layers", &
+    call set_axis_info(extra_axes(1), "cat", longname="Ice thickness categories", units="None", ax_size=IG%CatIce)
+    call set_axis_info(extra_axes(2), "cat0", longname="Ice thickness categories with open water", units="None", &
+                       ax_size=IG%CatIce+1)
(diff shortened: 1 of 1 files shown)


marshallward:
Seems inconsistent that `set_axis_info` has `units` as an optional argument, but is required by `get_file_fields`.   Maybe this is better resolved in the MOM_io and we should either redefine the argument as required, or have a default units value.

@Hallberg-NOAA Do you know why this was not an issue before?  Niki found this after using a nontrivial IO layout.

Hallberg-NOAA:
I agree with the diagnosis of where the model is failing, but adding meaningless units to (essentially) integer axes is not the right answer. Even if we were to add these units to files we create, it would not keep the reads from failing for other (decomposed) files created elsewhere.

  Rather the correct answer is to add a test for variable_att_exists() around every call for get_variable_attribute() in MOM6/config_src/infra/FMS2/MOM_io_infra.F90 so that it does not fail for files where these attributes are missing.

  I believe that this PR should _not_ be merged into SIS2, and should be closed. Do you agree @nikizadehgfdl ?

nikizadehgfdl:
OK.

PR #151, opened 2021-06-15
Added some documentation for Icepack ridging interface

MJHarrison-GFDL:


Hallberg-NOAA (changes requested):
I think that these comments reveal some problems that we likely have with the dimensional rescaling of the arguments to ridge_ice.  SIS2 variables are dimensionally rescaled, but ice_pack doesn't know anything about our clever approach for detecting dimensionally inconsistent expressions.  Therefore, all real arguments to ridge_ice need to be in MKS units.  Also, there is no such thing as the units for an integer or logical variable; such variables should not have units.

  I think that we need to go back and revisit the arguments that are being passed to ridge_ice, and undo the dimensional rescaling for SIS2 in ice_ridging in SIS2/src/ice_ridge.F90 when they are being converted to the variables that are passed to ridge_ice.  For this pull request, please just log the MKS units of the arguments, and we can deal with the other problem separately.

  I think that this exposed issue also reveals the need to figure out how to test a case where the icepack ridging is active in a SIS2 simulation.

Hallberg-NOAA (approved):

PR #154, opened 2021-07-15
+(*)Fixed the landfast ice algorithm

Hallberg-NOAA:
Revised the landfast ice discretization so that it is extracting momentum from
the ice and giving it to the land, rather than exchanging it with the ocean.
Also added the new diagnostics FLF_X and FLF_Y to capture the forces due to the
landfast ice, and applied the appropriate dimensional rescaling to the
parameters for the Lemieux landfast ice parameterization.  This will change
answers in cases that use LEMIEUX_LANDFAST = True, and it adds new entries to
the SIS.available_diags files, but all solutions in the MOM6-examples test suite
are bitwise identical.

### DIFF
--- a/src/SIS_dyn_cgrid.F90
+++ b/src/SIS_dyn_cgrid.F90
@@ -107,23 +107,23 @@ type, public :: SIS_C_dyn_CS ; private
   integer :: max_writes       !< The maximum number of times any PE can write out
                               !! a column's worth of accelerations during a run.
   logical :: lemieux_landfast !< If true, use the lemieux landfast ice parameterization.
-  real :: lemieux_k1          !< 1st free parameter for landfast parameterization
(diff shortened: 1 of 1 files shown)


kshedstrom:
I approve this PR. So wonderful to have the landfast ice slow down!

marshallward:
This PR has passed my implementation of the pipeline.

Two `SIS.available_diags` files require updates:
* `ice_ocean_SIS2/SIS2_bergs_cgrid/SIS.available_diags`
* `ice_ocean_SIS2/SIS2_cgrid/SIS.available_diags`

I will merge this now, and validate it in the MOM6 pipeline validation once we get that working.

kshedstrom:
```
  real :: ci_u ! Concentration and u-points [nondim]
  real :: ci_v ! Concentration and u-points [nondim]
```
Shouldn't this read "at" instead of "and"? Also, the second is v-points.

PR #156, opened 2021-09-01
+Change opt. arg. to set_massless_SIS_tracers

Hallberg-NOAA:
Changed the interface to set_massless_SIS_tracers by replacing the optional
compute_domain logical argument (which took the opposite sense of what its name
seems to imply) with a new optional integer argument, halos, to specify the
size of the halos that this works on.  Previously calls to this routine from
ice_cat_transport and finish_ice_transport had led to uninitialized data being
used, but this change fixes this problem.  All answers are bitwise identical,
although there is a change to a public interface.

### DIFF
--- a/src/SIS_tracer_registry.F90
+++ b/src/SIS_tracer_registry.F90
@@ -412,23 +412,24 @@ subroutine update_SIS_tracer_halos(TrReg, G, complete)
 end subroutine update_SIS_tracer_halos
 
 !> Set the properties of the ice tracers where there is no ice mass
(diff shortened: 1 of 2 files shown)


marshallward:
I don't see the SIS2 initialization errors anymore in the test that I am running, so I would presume that this is working correctly.

I will do a proper regression test next.

marshallward:
I verified this quite some time ago, but forgot to merge it. 

And I've inadvertently tested several more PRs under this branch.  Oops...

Merging now.

PR #159, opened 2021-11-19
MOM EOS pointer API update

marshallward:
This patch updates the equation of state (EOS) calls in SIS2 to use the
updated MOM EOS API:

* `EOS_end` is removed, since there is no longer anything to deallocate

* THE EOS type is allocated prior to calling `EOS_init`, rather than
  relying on the internal allocation.  This is because `EOS_init` no
  longer expects a pointer, and no longer does this allocation.

This patch may still work with the older version of MOM6, although the
EOS pointer will be double-allocated.

### DIFF
--- a/src/SIS2_ice_thm.F90
+++ b/src/SIS2_ice_thm.F90
@@ -4,7 +4,7 @@ module SIS2_ice_thm
 ! This file is part of SIS2. See LICENSE.md for the license.
 
 use ice_thm_mod,         only : get_thermo_coefs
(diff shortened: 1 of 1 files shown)


marshallward:
The companion MOM6 PR is here: https://github.com/NOAA-GFDL/MOM6/pull/5

Hallberg-NOAA:
This PR has been tested with the existing version of MOM6, and it works, reproducing all existing answers and output.

Careful inspection of the MOM6 code shows that there will be no double-allocation of EOS pointers.

PR #160, opened 2021-12-09
+Rescale G%Rad_Earth and set it via dG%Rad_Earth_L

Hallberg-NOAA:
Rescaled the units of the Rad_Earth element of SIS_hor_grid_type and set it
using the equivalently scale variable dG%Rad_Earth_L from the dyn_grid type.
This change led naturally to changed units for the I_area_Earth internal
variable in ice_diagnostics_init() and the cancellation of some grid rescaling
factors that were there before.  This commit requires that SIS2 use a recent
(November 27, 2021 or later) version of MOM6 from the dev/gfdl branch, and will
not compile with earlier versions of the MOM6 code.  All answers are bitwise
identical, but there is one minor interface change in a widely used transparent
type.


marshallward:
This has passed local regression, and will be verified in the MOM6 pipeline.

PR #162, opened 2021-12-21
Probabilistic landfast ice from CICE

kshedstrom:
A different way to do landfast ice, unpublished as yet by the Canadians.

### DIFF
--- a/src/SIS_continuity.F90
+++ b/src/SIS_continuity.F90
@@ -1204,7 +1204,7 @@ subroutine meridional_mass_flux(v, dt, G, US, IG, CS, LB, h_in, vh, htot_in, vh_
     I_htot, &  ! The inverse of htot or 0 [R-1 Z-1 ~> m2 kg-1].
     hl, hr     ! Left and right face thicknesses [R Z ~> kg m-2].
   real, dimension(SZI_(G)) :: &
-    vhtot      ! The total transports [R Z L2 s-1 ~> kg s-1].
+    vhtot      ! The total transports [R Z L2 T-1 ~> kg s-1].
   real :: CFL ! The CFL number based on the local velocity and grid spacing [nondim].
   real :: curv_3 ! A measure of the thickness curvature over a grid length,
                  ! with the same units as h_in.
@@ -1323,7 +1323,7 @@ subroutine meridional_mass_flux(v, dt, G, US, IG, CS, LB, h_in, vh, htot_in, vh_
       enddo
     elseif (present(vh_tot)) then
       do i=ish,ieh
-        vh_tot(i,J) = vhtot(I)
+        vh_tot(i,J) = vhtot(i)
       enddo
     endif
 
--- a/src/SIS_continuity.F90
+++ b/src/SIS_continuity.F90
@@ -102,9 +102,10 @@ subroutine ice_continuity(u, v, hin, h, uh, vh, dt, G, US, IG, CS, use_h_neg, ma
   type(loop_bounds_type) :: LB  ! A structure with the active loop bounds.
   real    :: h_up ! The upwind thickness [R Z ~> kg m-2]
   logical :: apply_h_neg
-  integer :: is, ie, js, je, nCat, stensil
+  integer :: is, ie, js, je, nCat, stencil
   integer :: i, j, k
 
+  character(len=256) :: mesg
   logical :: x_first
   is = G%isc ; ie = G%iec ; js = G%jsc ; je = G%jec ; nCat = IG%CatIce
 
@@ -117,10 +118,11 @@ subroutine ice_continuity(u, v, hin, h, uh, vh, dt, G, US, IG, CS, use_h_neg, ma
     apply_h_neg = ((use_h_neg) .and. (CS%h_neglect_cont > 0.0))
   endif
 
-  stensil = 3 ; if (CS%simple_2nd) stensil = 2 ; if (CS%upwind_1st) stensil = 1
+  stencil = 3 ; if (CS%simple_2nd) stencil = 2 ; if (CS%upwind_1st) stencil = 1
 
   do k=1,nCat ; do j=js,je ; do i=is,ie ; if (h(i,j,k) < 0.0) then
-    print *, 'Negative ice thickness at:', i+G%idg_offset, j+G%jdg_offset, k, h(i,j,:)
+    write(mesg,'("Negative ice thickness at: ", 3i6, 1pe12.4)') i+G%idg_offset, j+G%jdg_offset, k, h(i,j,k)
+    call SIS_error(WARNING, mesg, all_print=.true.)
     call SIS_error(FATAL, 'Negative thickness input to ice_continuity().')
   endif ; enddo ; enddo ; enddo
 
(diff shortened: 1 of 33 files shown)


src/SIS_dyn_cgrid.F90
```diff
@@ -1737,17 +1748,183 @@ subroutine basal_stress_coeff_C(G, mi, ci, sea_lev, CS)
...
+
+  do j=jsc-1,jec+1
+    do i=isc-1,iec+1
+
+      acat(1:ncat) = IST%part_size(i,j,1:ncat)
+      atot = sum(acat)
+      vcat(1:ncat) = IST%mH_ice(i,j,1:ncat) / rho_ice
+      m_i = sum(vcat)
+
+      if (atot > 0.05 .and. m_i > 0.01 .and. sea_lev(i,j) < max_depth) then
```
Hallberg-NOAA, line 1851: There are dimensionally inconsistent comparisons and hard-coded dimenesional constansts here. Both m_i and sea_lev have units here of [Z ~> m], while max_depth is a hard-coded number with units of [m], and 0.01 is presumbly 0.01 [m].  Please correct the units, and verify that the rescaling is correct by trying out a test case with various values Z_RESCALE_POW and R_RESCALE_POW.

src/SIS_dyn_cgrid.F90
```diff
@@ -1737,17 +1748,183 @@ subroutine basal_stress_coeff_C(G, mi, ci, sea_lev, CS)
...
+
+  integer, dimension(ncat_b) :: tmp
+  real, dimension(0:IG%CatIce) :: hin_max   ! category limits (m)
+
+  logical, dimension (ncat_b) :: gt
+
+  real, dimension(IG%CatIce) :: vcat   ! category limits (m)
+  real, dimension(IG%CatIce) :: acat   ! category limits (m)
+
+  real :: wid_i, wid_b, mu_i, sigma_i, mu_b, m_i, v_i !  parameters for PDFs
```
Hallberg-NOAA, line 1807: Please add comments describing the purpose and units of each of these real variables.

src/SIS_continuity.F90
```diff
@@ -120,6 +120,7 @@ subroutine ice_continuity(u, v, hin, h, uh, vh, dt, G, US, IG, CS, use_h_neg, ma
   stensil = 3 ; if (CS%simple_2nd) stensil = 2 ; if (CS%upwind_1st) stensil = 1
 
   do k=1,nCat ; do j=js,je ; do i=is,ie ; if (h(i,j,k) < 0.0) then
+    print *, 'Negative ice thickness at:', i+G%idg_offset, j+G%jdg_offset, k, h(i,j,:)
```
Hallberg-NOAA, line 123: We have been encouraged not to write directly to stdout using `write (*,*)` or `print *` to prepare for machines where not all PEs are able to access the I/O channels, and we are not doing so elsewhere in the MOM6 or SIS2 code.

  Please revise these useful messages to write to a message string, and then either concatenate this message string to the following `SIS_error(FATAL, ...)` call, or as 'call SIS_error(WARNING, mesg, all_print=.true.)` before the fatal error message, and similarly for the other 5 new error messages in this file and the other new error message in SIS_transport.F90 using the similar `print *,...` construct.

Hallberg-NOAA (changes requested):
Overall, this PR looks like a valuable new contribution, but I believe that there are some expressions that do not have the appropriate dimensional consistency verification and there are some error messages that do not follow the I/O pattern we have been advised to use.  Please consider making the changes suggested in the specific comments on parts of the code involved in this PR.

kshedstrom:
Thank you for all your suggestions!

1. For the scalings, do you do MOM_override and SIS_override together or independently? Either should work? Both Z and R rescalings are problematic for me. Do you have debugging hints?

2. I have gotten rid of the print statements, however the following compiles, but doesn't run:
```
write(mesg,'("Negative ice volume after ridging: ", 2i6, 2x, (1pe12.4))') i+G%jdg_offset, j+G%jdg_offset, aicen, vicen
```

kshedstrom:
Thanks, yes, I was getting a mismatch with my illegal print statements.

src/SIS_dyn_cgrid.F90
```diff
@@ -1738,17 +1781,192 @@ subroutine basal_stress_coeff_C(G, mi, ci, sea_lev, CS)
...
+subroutine basal_stress_coeff_itd(G, IG, IST, sea_lev, CS)
+
+  type(SIS_hor_grid_type),            intent(in)  :: G     !< The horizontal grid type
+  type(ice_grid_type),                intent(in)  :: IG    !< The sea-ice specific grid type
+  type(ice_state_type),               intent(in)  :: IST   !< A type describing the state of the sea ice
+  real, dimension(SZI_(G),SZJ_(G)),   intent(in)  :: sea_lev !< Sea level [Z ~> m]
+  type(SIS_C_dyn_CS),                 pointer     :: CS    !< The control structure for this module
+
+  integer, parameter :: &
+           ncat_b = 100, &  ! number of bathymetry categories
```
Hallberg-NOAA, line 1815: Given that there are some operations that occur ncat_b x ncat_i times for each grid cell, it seems dangerous to specify these as (large) fixed parameters.  It is hard-coded that this routine will do 10,000 operations at every grid point that is both shallow and has enough ice.  It seems to me that these sizes should be model variables that can be determined and set at run-time.

src/SIS_dyn_cgrid.F90
```diff
@@ -1738,17 +1781,192 @@ subroutine basal_stress_coeff_C(G, mi, ci, sea_lev, CS)
...
+
+        P_x(:) = g_k(:) * wid_i
+        P_y(:) = b_n(:) * wid_b
+
+        do n =1, ncat_i
+          if (x_k(n) > x_kmax) P_x(n)=0.0
+        enddo
+
+        ! calculate Tb factor at t-location
+        do n=1, ncat_i
```
Hallberg-NOAA, line 1939: This line is hard-coded to run 10,000 times for every model grid cell.  This seems inflexible and possibly excessive?  I think that the values of ncat_i and ncat_b should be something that can be set at run-time.

Hallberg-NOAA (commented):
I have now read through this PR, and I agree that most of this seems like the right thing to do.  However, I do not like that the sizes of the ice and bathymetry distributions are both set at 100 elements, leading to a 10,000 step operation at some points.  I think that these sizes should be runtime configurable parameters, and not hard-coded.

Also, I do not like that this new ITD_LANDFAST capability by design does not work with MERGED_CONTINUITY, which is supposed to be the direction we are going with our ice-ocean coupling.  Is there a fundamental and irreconcilable conflict, or is the error message just a placeholder until we get the chance to try to get these two options working together?

kshedstrom:
I had to go digging to remind myself why I didn't think it worked with MERGED_CONTINUITY. The Lemieux scheme is called from either SIS_merged_dyn_cont or from SIS_dynamics_trans. The ITD scheme needs the ice thickness distribution which is not currently passed to SIS_merged_dyn_cont since it only wants the simplified state. Would you object to having it know about the ITD just for this?

Yes, I agree that 10,000 iterations seems like a lot. I will add some more runtime parameters.

Hallberg-NOAA (approved):
These changes have addressed the concerns that I had with the previous version.

PR #163, opened 2021-12-29
+(*)Report heat stocks relative to liquid at 0 degC

Hallberg-NOAA:
Revised ice_stock_pe() to report the heat stock relative to liquid water at 0
degrees C, rather than just the heat required to melt the ice, as this gives a
much more robust estimates of the enthalpy budget when sea ice salinities are
allowed to vary.  To help implement this, the new function energy_0degC was
added to SIS2_ice_thm.F90.  Several module use statements that were referring to
unused routines from within SIS2_ice_thm were also redacted.  All solutions are
bitwise identical, but there are changes in the reported stock values, and there
is a new publicly visible function.


MJHarrison-GFDL (approved):
In addition to some code cleanup and documentation, this change will allow for a more accurate accounting of enthalpy exchange between the sea-ice and ocean.  Answers are not impacted, but as noted the stocks report (which is used within the FMS coupling framework) will change but these results are not being tracked in current MOM6 regression tests

Issue #167, opened 2022-02-07
Issue compiling: src/SIS_sum_output.F90


uramirez8707:
I get the following compile error when trying to compile the dev/gfdl branch of SIS2:

```
src/SIS_sum_output.F90(174): error #6460: This is not a field name that is defined in the encompassing structure.   [WATER_CELL_PREV]
    allocate(CS%water_cell_prev(G%isd:G%ied, G%jsd:G%jed), source=0.0)
```

and similar errors for `CS%heat_cell_prev` and `CS%salt_cell_prev`.

CS is `type(SIS_sum_out_CS`
https://github.com/NOAA-GFDL/SIS2/blob/87ff9e51982f8c930db9720274b88cec6d4a2b39/src/SIS_sum_output.F90#L104-L105

Which does not have the `water_cell_prev ` `heat_cell_prev` and `salt_cell_prev` defined. 
https://github.com/NOAA-GFDL/SIS2/blob/dev/gfdl/src/SIS_sum_output.F90#L47-L93

Renaming `*_col_prev` to `*_cell_prev` in the derived type solves the issue, but I don't know if that is the right thing to do.

marshallward:
I've reproduced this error.  I'll chat with the others and get back to you soon.

PR #168, opened 2022-02-08
Fix the result of a bad automated merge

Hallberg-NOAA:
In merging in two separate pull requests that made corrections to similar
lines in the same file, three renamed variables retained their original names,
resulting in SIS2 code that does not compile.  Each of the PRs was fully tested
separately, but for whatever reason their combination was not, and the SIS2 git
repository does not seem to detect certain merge conflicts.  This PR updates
these variable names to their intended values, and the code is once more
compiling and is reproducing the intended outputs.

  This PR will address SIS2 issue #167, which can be closed once this PR is accepted.


marshallward:
This has passed my internal regression, so I will merge it.

I will verify this PR in the MOM6 regression suite.

PR #171, opened 2022-03-22
Fix Bad ice warnings over land

nikizadehgfdl:
- CM4/ESM4 coupled models get loads of warnings in stdout (causing 2G
fms.out files per year) like:
WARNING from PE   704: Bad ice data End set_ice_surface_state ;  at   20.2   6.5 or i,j,k =    5   5   0; nbad =    240 on pe  704 ; T_sfc =   0.0000E+00, ps =   1.0000E+00, S_sfc =   0.0000E+00
- This is because t_surf is set to 0 after allocation and then compared
with a minimum 273.15-100 (K) to catch bad surface values over sea ice
but the comparison is not restricted to ocean and is causing warnings
over land in fully coupled models.
- The fix is to restrinct the comparison over ocean mask just like is
done in SIS_types.F90

### DIFF
--- a/src/ice_type.F90
+++ b/src/ice_type.F90
@@ -455,25 +455,25 @@ subroutine Ice_public_type_bounds_check(Ice, G, msg)
   n_bad = 0 ; i_bad = 0 ; j_bad = 0 ; k_bad = 0
 
   t_min = T_0degC-100. ; t_max = T_0degC+60.
-  do k=0,ncat ; do j=jsc,jec ; do i=isc,iec
+  do k=0,ncat ; do j=jsc,jec ; do i=isc,iec ; if (G%mask2dT(i,j)>0.5) then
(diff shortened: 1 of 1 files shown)


MJHarrison-GFDL:
I approve of this PR.  This patch will get rid of irrelevant warning messages for the first and subsequent model segments.

NOTE: It is not clear why we are using a mask2dT>0.5 conditon rather than mask2dT>0.0 , but this is the  (mostly) standardized way to filter land points in SIS2. In practice, mask2dT is either 0 or 1. - if the somewhat confusing APPLY_MASKS_AFTER_RESTART flag is set to True, then land values are reset to 0C (instead of 0K) after restarts.

marshallward:
Gaea regression: https://gitlab.gfdl.noaa.gov/ogrp/SIS2/-/pipelines/15165 :heavy_check_mark:

Issue #169, opened 2022-02-26
Doesn't get stresses right at OBCs


kshedstrom:
In going from stresses at h points to stresses on the u/v points, it counts on G%masks2dCu to block exterior values, but that is 1 on open boundaries. We need some OBC code to clean up the mess afterwards.

PR #173, opened 2022-04-09
Barest start of ice OBCs

kshedstrom:
Copying code from MOM6 for parsing the OBC strings. This is enough to solve #169 and to allow perfect restarts with MERGED_CONTINUITY. Depends on a pending PR in MOM6 and on adding the simplest OBC stuff into SIS_input so that the SIS2 and MOM6 OBC locations match.


Hallberg-NOAA:
There don't seem to be any code changes to the dev/gfdl code associated with this pull request.  Is there some mistake here?

kshedstrom:
Sorry, stumbling on git again.

PR #175, opened 2022-05-22
+(*)Write unscaled SIS2 restart files

Hallberg-NOAA:
Revised the SIS2 code so that the output restart files are always exactly the
same as they would be if no dimensional rescaling is applied.  The input restart
files can still have rescaling, so files written by previous versions of the
code still work exactly as before.  This does change the output, in the sense
that the restart files are unscaled and some units documents in the restart
files are corrected, but the solutions themselves are bitwise identical. Also,
there are new (non-optional) unit scaling type arguments to several routines.

The specific changes in this commit include:

 - Add a new optional conversion arguments to all of the register_restart_field
   routines in SIS_restart.F90.

 - Add a new optional scale argument to all of the only_read_restart_field
   routines in SIS_restart.F90.

 - Added conversion factors to all register_restart_field() or
   register_restart_pair() calls for variables that are subject to dimensional
   rescaling.

 - Revised the calculation of restart rescaling factors to reflect the rescaling
   that now occurs directly in the restore_state() call.

 - Added new US arguments to SIS_C_dyn_register_restarts(),
   SIS_B_dyn_register_restarts(), SIS_dyn_trans_register_restarts(),
   ice_state_register_restarts(), ice_state_read_alt_restarts() and
   register_fast_to_slow_restarts()

 - Corrected the rescaling of hin_max in ice_ridging().

 - IG%H_to_kg_m2 is no longer being used anywhere, but rather than just
   eliminating it, it is being retained in the code to maintain compatibility
   with older restart files.

 - Removed the unused H_to_kg_m2 argument to rescale_ice_state_restart_fields()

### DIFF
--- a/src/SIS_dyn_bgrid.F90
+++ b/src/SIS_dyn_bgrid.F90
@@ -689,11 +689,12 @@ end function sigII
 !~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~!
 !> SIS_B_dyn_register_restarts allocates and registers any variables for this
 !!      module that need to be included in the restart files.
-subroutine SIS_B_dyn_register_restarts(HI, param_file, CS, Ice_restart)
+subroutine SIS_B_dyn_register_restarts(HI, param_file, CS, US, Ice_restart)
   type(hor_index_type),    intent(in) :: HI    !< The horizontal index type describing the domain
   type(param_file_type),   intent(in) :: param_file !< A structure to parse for run-time parameters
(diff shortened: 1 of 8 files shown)


adcroft (changes requested):
Test https://gitlab.gfdl.noaa.gov/ogrp/SIS2/-/pipelines/15679 passed BUT a rebase merge has conflicts. Please rebase on dev/gfdl and force push your branch.

Hallberg-NOAA:
I have now done the requested rebase to dev/gfdl and force push to my branch.

adcroft (approved):
Re-test https://gitlab.gfdl.noaa.gov/ogrp/SIS2/-/pipelines/15689 passed.

PR #176, opened 2022-05-24
Replace MOM_memory.h with SIS2_memory.h

marshallward:
The most recent PR to SIS2 includes MOM_memory.h in one of its files,
which breaks certain SIS2-only library builds.

This patch replaces it with SIS2_memory.h, which is functionally
identical but plays better with existing builds.

### DIFF
--- a/src/SIS_open_boundary.F90
+++ b/src/SIS_open_boundary.F90
@@ -19,7 +19,7 @@ use MOM_unit_scaling,         only : unit_scale_type
 
 implicit none ; private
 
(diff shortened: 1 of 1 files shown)


marshallward:
Gaea regression: https://gitlab.gfdl.noaa.gov/ogrp/SIS2/-/pipelines/15600

marshallward:
Given that the current MOM6 CI is failing without this change, and that we've already discussed this in other channels, I will break protocol and merge this.

PR #179, opened 2022-08-03
Icepack bugfix sst

MJHarrison-GFDL:
This change fixes a problem with initialization of sea surface temperature information for the Icepack module and should avoid uninitalized array access within Icepack.  This should not impact answers for existing simulations using Icepack ridging.


Hallberg-NOAA (approved):
I agree that these changes are correct, and moreover that they will address the issue at the source of https://github.com/mom-ocean/MOM6/issues/1461, which can now be closed.

I have run these changes through the suite SIS2-based models in the MOM6-examples test suite, and all answers are bitwise identical.  I could envision that there might be answer changes in cases using the Icepack-derived ridging scheme, but if so I am confident that the new answers are the right ones, and that they avoid the possibility of NaNs in uninitialized arrays bringing down the model, and hence that there are no defensible previous answers that could be preserved.

marshallward:
Gaea regression: https://gitlab.gfdl.noaa.gov/ogrp/SIS2/-/pipelines/16281 :heavy_check_mark:

Issue #183, opened 2022-09-20
Unhelpful error message if asking for DO_RIDGING without Icepack linked in


kshedstrom:
If asking for DO_RIDGING without Icepack, one gets:
```
FATAL from PE    14: nilyr or nslyr mismatch with Icepack
```
It would be nice to having it instead say that DO_RIDGING is not available without linking in Icepack/columnphysics.

MJHarrison-GFDL:
Thanks for pointing this out @kshedstrom .  It is an accurate, but useless error message.  I will send a PR with a more useful message informing that actual Icepack code needs to be included in the compilation of the model, rather than the stub code in config_src/external

Hallberg-NOAA:
This issue was addressed by https://github.com/NOAA-GFDL/SIS2/pull/184, and can now be closed.

Issue #178, opened 2022-07-28
New ridging options not all that stable in SIS2


kshedstrom:
I bring this up because the old ridging options are being obsoleted in Icepack. In my experience, both NEW_RIDGE_REDISTRIBUTION and NEW_RIDGE_PARTICIPATION can lead to random blow-ups. For example:
```
WARNING from PE     4: Negative ice volume after ridging:    181    13    9.2766E-57 -7.3863E-67
FATAL from PE     4: Input to adjust_ice_categories, non-zero snow mass rests atop no ice.
```

kshedstrom:
I talked them out of obsoleting it for now.

MitchBushuk:
Thank you Kate, much appreciated. I have also been running into the same issue in SIS2/MOM6 experiments, generally with the following error:
WARNING from PE   513: Negative ice thickness at:     92   237     4 -1.0998E-65
FATAL from PE   513: Negative thickness input to ice_continuity().

I have found that the model runs successfully with NEW_RIDGE_PARTICIPATION=true and NEW_RIDGE_REDISTRIBUTION=false. But any configuration with NEW_RIDGE_REDISTRIBUTION=true has been crashing due to tiny negative thickness values. I tried setting the CONTINUITY_H_NEGLECT parameter to a small positive value (1e-20), but this did not prevent the error. I also tried some different options for SIS_THICKNESS_ADVECTION_SCHEME and SIS_CONTINUITY_SCHEME but those also didn't seem to help.

kshedstrom:
In my code, there's a max function so that vicen coming out of the ridging can't be negative. Then I've got an ice category with no ice, but some remnant snow. That should be fixable.

Yes, one of those options is more troublesome than the other, but I did have them both independently go bad on me at least once.

kshedstrom:
I added this to the if block on negative ice:
```
            vicen(k) = 0.0 
            vsnon(k) = 0.0
```
With both:
```
#override NEW_RIDGE_REDISTRIBUTION = True
#override NEW_RIDGE_PARTICIPATION = True
```
My Bering domain runs through the winter (with plenty of warnings), then dies in June with:
```
FATAL from PE    22: Negative thickness input to ice_continuity().
```

MJHarrison-GFDL:
[...]

PR #184, opened 2022-09-29
Icepack area underflow patch 

MJHarrison-GFDL:
New parameter RIDGE_AREA_UNDERFLOW is introduced in order to avoid the occurrence of undetectably small fractional area and negative ice volume following ridging adjustments from Icepack. The default is 0.0, which does not introduce changes in existing runs.  A reasonable value would be 10^-26 which would correspond to an Angstrom-scale ice patch within a km-scale grid cell.  This addresses #178 and #183

### DIFF
--- a/src/ice_ridge.F90
+++ b/src/ice_ridge.F90
@@ -50,7 +50,7 @@ type, public :: ice_ridging_CS ; private
   real :: mu_rdg = 3.0 !< e-folding scale of ridged ice, new_rdg_partic (m^0.5)
   real :: area_underflow = 0.0 ! a non-dimesional fractional area underflow limit for the sea-ice
                       ! ridging scheme. This is defaulted to zero, but a reasonable value might be
-                      ! 10^-24 which for a km square grid cell would equate to an Angstrom scale ice patch.
+                      ! 10^-26 which for a km square grid cell would equate to an Angstrom scale ice patch.
 end type ice_ridging_CS
 
 contains
@@ -79,7 +79,7 @@ subroutine ice_ridging_init(G, IG, PF, CS, US)
                    units="m^0.5", default=3.0)
     call get_param(PF, mdl, "RIDGE_AREA_UNDERFLOW", CS%area_underflow, &
                    "A fractional area limit below which ice fraction is set to zero "//&
-                   "A reasonable default value for a km scale grid cell is 10^-32.",&
+                   "A reasonable default value for a km scale grid cell is 10^-24.",&
                    units="none", default=0.0)
   endif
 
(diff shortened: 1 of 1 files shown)


MitchBushuk (approved):
Thanks @MJHarrison-GFDL for putting this ice area underflow fix in.

MitchBushuk (approved):

src/ice_ridge.F90
```diff
@@ -43,9 +45,12 @@ module ice_ridging_mod
 
 type, public :: ice_ridging_CS ; private
   logical :: &
-       new_rdg_partic = .false., & !< .true. = new participation, .false. = Thorndike et al 75
-       new_rdg_redist = .false.    !< .true. = new redistribution, .false. = Hibler 80
+  new_rdg_partic = .false., & !< .true. = new participation, .false. = Thorndike et al 75
+  new_rdg_redist = .false.    !< .true. = new redistribution, .false. = Hibler 80
   real :: mu_rdg = 3.0 !< e-folding scale of ridged ice, new_rdg_partic (m^0.5)
+  real :: area_underflow = 0.0 ! a non-dimesional fractional area underflow limit for the sea-ice
```
marshallward, line 51: @MJHarrison-GFDL Could you rewrite this in doxygen syntax?

src/ice_ridge.F90
```diff
@@ -239,7 +248,11 @@ subroutine ice_ridging(IST, G, IG, mca_ice, mca_snow, mca_pond, TrReg, CS, US, d
 
   call icepack_query_tracer_sizes(ncat_out=ncat_out,ntrcr_out=ntrcr_out, nilyr_out=nilyr_out, nslyr_out=nslyr_out)
 
-  if (nIlyr .ne. nilyr_out .or. nSlyr .ne. nslyr_out ) call SIS_error(FATAL,'nilyr or nslyr mismatch with Icepack')
+  if (nIlyr .ne. nilyr_out .or. nSlyr .ne. nslyr_out ) &
```
marshallward, line 251: This indentation block needs to be fixed.
(We generally require >2 space indentation for line continuations.  The one exception is after an if-block, which is only 2 spaces.   So the first line should be 2 spaces, not one, and the other lines should be "more than 2" (your discretion).

src/ice_ridge.F90
marshallward: Actually these indentations also need to be fixed.  ("More than 2 after line continuation")

marshallward:
Gaea regression: https://gitlab.gfdl.noaa.gov/ogrp/SIS2/-/pipelines/16960 :heavy_check_mark: 

I will squash-merge this PR.

marshallward (approved):

kshedstrom:
I'm afraid you have to check for negative ice volume as well. I get:
```
WARNING from PE    63: Negative ice thickness at:    233   235     7 -6.2085E-61

FATAL from PE    63: Negative thickness input to ice_continuity().
```

PR #185, opened 2022-11-17
Ponding

kshedstrom:
This is incomplete, but I thought I'd get it out there for feedback. I'm planning to call the Icepack routines for the ponds, but there's work to do on the thermal properties too.

### DIFF
--- a/src/SIS_transport.F90
+++ b/src/SIS_transport.F90
@@ -81,9 +81,11 @@ end type SIS_transport_CS
 type, public :: cell_average_state_type ; private
   real, allocatable, dimension(:,:,:) :: m_ice  !< The mass of ice in each thickness category
                                                 !! per unit total area in a cell [R Z ~> kg m-2].
-  real, allocatable, dimension(:,:,:) :: m_snow !< The mass of ice in each thickness category
+  real, allocatable, dimension(:,:,:) :: m_snow !< The mass of snow in each thickness category
                                                 !! per unit total area in a cell [R Z ~> kg m-2].
   real, allocatable, dimension(:,:,:) :: m_pond !< The mass of melt pond water in each thickness
+                                                !! category per unit total area in a cell [R Z ~> kg m-2].
+  real, allocatable, dimension(:,:,:) :: m_pond_ice !< The mass of melt pond ice in each thickness
                                                 !! category per unit total area in a cell [R Z ~> kg m-2].
   real, allocatable, dimension(:,:,:) :: mH_ice !< The mass of ice in each thickness category
                                                 !! per unit of ice area in a cell [R Z ~> kg m-2].  The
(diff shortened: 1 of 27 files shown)


src/SIS_transport.F90
```diff
@@ -208,10 +219,12 @@ subroutine ice_cat_transport(CAS, TrReg, dt_slow, nsteps, G, US, IG, CS, uc, vc,
     ! Accumulated diagnostics
     CAS%dt_sum = CAS%dt_sum + dt_adv
     if (allocated(CAS%uh_sum)) then ; do k=1,nCat ; do j=jsc,jec ; do I=isc-1,iec
-      CAS%uh_sum(I,j) = CAS%uh_sum(I,j) + dt_adv * ((uh_pond(I,j,k) + uh_snow(I,j,k)) + uh_ice(I,j,k))
+      CAS%uh_sum(I,j) = CAS%uh_sum(I,j) + dt_adv * ((uh_pond(I,j,k) + uh_snow(I,j,k)) + uh_ice(I,j,k) + \
```
Hallberg-NOAA, line 222: Both in this expression for `CAS%uh_sum` and in the subsequent one for `CAS%vh_sum`, we should add parentheses to specify the order of the sums.  As written, the compiler is free to choose the order with which `uh_pond_ice` is included in the sum.  My personal preference would be to combine the `uh_pond` and `uh_pond_ice` first, but what is more important is that we force an order to the sums to help ensure that we get the same answers when compiled for debugging or for optimized runs. 

Also, the use of `\` as the continuation character is not found elsewhere in the SIS2 code.  Please use `&` for consistency.

src/ice_pond.F90
```diff
@@ -0,0 +1,513 @@
...
+
+  if (.not.associated(CS)) allocate(CS)
+! call get_param(PF, mdl, "NEW_RIDGE_PARTICIPATION", CS%new_rdg_partic, &
+!                "Participation function used in ponds, .false. for Thorndike et al. 1975 "//&
+!                ".true. for Lipscomb et al. 2007", default=.false.)
+! call get_param(PF, mdl, "NEW_RIDGE_REDISTRIBUTION", CS%new_rdg_redist, &
+!                "Redistribution function used in ponds, .false. for Hibler 1980 "//&
+!                ".true. for Lipscomb et al. 2007", default=.false.)
+
+  ncat=IG%CatIce ! The number of sea-ice thickness categories
```
Hallberg-NOAA, line 72: Please use white space around the equal signs here, as in `nilyr = IG%NkIce`.  As discussed in the MOM6 style guide at https://github.com/NOAA-GFDL/MOM6/wiki/Code-style-guide, which also pertains to SIS2, we reserve the use of the equal sign without spaces for named arguments, as in `default=.false.` two lines above.  This is a minor stylistic thing, but it does help when searching through the code for irregular patterns that might be bugs, and we think that it promotes the human readability of the code.

Hallberg-NOAA (changes requested):
Overall this seems like a well-conceived and very valuable new contribution.  Thank you, @kshedstrom !

  I have a handful of minor stylistic comments that you might consider (see the other comments on this PR), but once these minor changes are made (or the suggestions refuted), I think that this will be ready to go.

kshedstrom:
Thanks for the feedback! Fixed, I think, also some fixes to the ridging routine I copied the bad habits from.

Hallberg-NOAA:
The only outstanding issues here are the stylistic changes to follow the MOM6 style guide.  Rather than hold this up any longer, I am going to approve this with the understanding that there may be code style cleanup as a part of subsequent commits.

Hallberg-NOAA (approved):

Hallberg-NOAA (changes requested):
Unfortunately because this PR is using the CVmix interfaces directly, it is failing to compile in our pipeline tests.  To address this problem, we need to add stubs of the icepack modules that we are using to config_src/external/Icepack_interfaces, similarly to the other files that are already in that are already in that directory.

PR #189, opened 2022-12-02
Fix for sigma_i going to zero in ITD landfast ice.

kshedstrom:
- We can't divide by zero, so don't let sigma_i be zero.

### DIFF
--- a/src/SIS_dyn_cgrid.F90
+++ b/src/SIS_dyn_cgrid.F90
@@ -1975,6 +1975,7 @@ subroutine basal_stress_coeff_itd(G, IG, IST, sea_lev, CS)
         enddo
         x_kmax = min(cut, x_kmax)
 
+        sigma_i = max(sigma_i, CS%puny)
         g_k(:) = exp(-(log(x_k(:)/CS%onemeter) - mu_i) ** 2 / (2.0 * sigma_i ** 2)) / &
                  (x_k(:) * sigma_i * sqrt(2.0 * pi))
 
--- a/src/SIS_dyn_cgrid.F90
+++ b/src/SIS_dyn_cgrid.F90
@@ -1952,16 +1952,11 @@ subroutine basal_stress_coeff_itd(G, IG, IST, sea_lev, CS)
         do n =1, ncat
           v_i = v_i + vcat(n)**2 / (max(acat(n), CS%puny))
         enddo
-        v_i = v_i - m_i**2
+        v_i = max( (v_i - m_i**2), CS%puny )
 
         ! parameters for the log-normal
         mu_i    = log(m_i/(CS%onemeter * sqrt(1.0 + v_i/m_i**2)))
-        sigma_i = 0.0
-        if (v_i > 0.0) then
-          sigma_i = max(sqrt(log(1.0 + v_i/m_i**2)), CS%puny)
-        else
-          sigma_i = CS%puny
-        endif
+        sigma_i = max(sqrt(log(1.0 + v_i/m_i**2)), CS%puny)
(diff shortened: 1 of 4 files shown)


Hallberg-NOAA (commented):
I do not understand why this change is necessary.   When I look at the code as a whole, I see that the expression for sigma_i about 20 lines above this change, `sigma_i = max(sqrt(log(1.0 + v_i/m_i**2)), CS%puny)` at line 1959 of this same file, already includes a max with `CS%puny`, and I am not seeing anythng between these lines that changes the value of `sigma_i`.  What am I missing here?

kshedstrom:
Yes, you are right. I don't remember how I thought sigma_i was getting a zero, except that it was thin new ice.

kshedstrom:
Now I remember - thin new ice lead to v_i being zero.

kshedstrom:
Sorry, @Hallberg-NOAA, getting my branches confused. How about this one?

Hallberg-NOAA (approved):
This change makes much more sense than the previous version!

src/SIS_dyn_cgrid.F90
```diff
@@ -1952,14 +1953,14 @@ subroutine basal_stress_coeff_itd(G, IG, IST, sea_lev, CS)
         do n =1, ncat
           v_i = v_i + vcat(n)**2 / (max(acat(n), CS%puny))
         enddo
-        v_i = v_i - m_i**2
+        v_i = max( (v_i - m_i**2), CS%puny )
```
Hallberg-NOAA, line 1956: I think that this expression is not dimensionally consistent. CS%puny is described as being non-dimensional (and is used as such on line 1960, but here it is being used in a expression with units of [m2].  Should there be a separate CS%puny_area variable?

marshallward:
@kshedstrom Could you do something like this?
```
if (sigma_i > 0) then
   g_k(:) = exp(-(log(x_k(:)/CS%onemeter) - mu_i) ** 2 / (2.0 * sigma_i ** 2)) / &
                 (x_k(:) * sigma_i * sqrt(2.0 * pi))
                 
   ...
   Tbt(:) = ...
else
   ! g_k(:) = 0., since exp(-a**2/n**2)/n -> 0 as n->0
   ! P_x(:) = 0.
   ! tb_tmp(:) = 0.

   Tbt(:) = 0.
endif
```
It would avoid the need for `CS%puny` here, and might also reduce the computation.

marshallward:
Thanks for testing it out, let's see what happens :sweat_smile:.

However, I think the better test would be to remove `puny_m2`, rather than introducing it alongside the `sigma_i > 0` test.  `sigma_i` is only zero if `v_i` is zero, so if you bound `v_i` then I would assume that it's not even possible for `sigma_i` to be zero.

I was thinking that the `sigma_i > 0` test would avoid the need to introduce another "puny" parameter, but I suppose that either approach gets you to the same place.

kshedstrom:
Try that one?

marshallward:
That's what I had in mind, I guess the question is whether it fixed your problem without creating any new ones?

kshedstrom:
My short job ran fine, a longer one is stuck in the queue.

kshedstrom:
Yes, the year-long job finished successfully!

marshallward:
Gaea regression: https://gitlab.gfdl.noaa.gov/ogrp/SIS2/-/pipelines/19489 :heavy_check_mark:

marshallward:
@kshedstrom Can I squash-merge this one?

kshedstrom:
Yes, of course.

PR #194, opened 2023-04-14
Correct interspersed option in the CIOD

theresa-morrison:
Changes to the combined ice-ocean driver (CIOD) to fix a bug. The state of the ocean is now passed to the ice through the OIB (ocean ice boundary type). A new subroutine has been added to combined_ice_ocean_driver.F90: direct_flux_ocn_to_OIB which is analogous to the existing direct_flux_ice_to_IOB subroutine. The new subroutine assumes the ocean and ice are on the same model grid and that the do_area_weighted_flux option is false for the frazil ice.

This new version of the CIOD requires the OIB as an input when update_slow_ice_and_ocean is called. Currently it is an optional argument, but there is a fatal error if the CIOD is used without it. Therefore, to use the corrected version of the CIOD the call to update_slow_ice_and_ocean in coupler_main must be modified.

The ice_model_init routine is changed in two ways. First, the concurrent atmosphere and concurrent ice options are now both inputs so they are no longer assumed to be the same. Second, there are minor changes to the slow and fast processor calls so the slow/fast processors can be shared. 

The direct_flux_ocn_to_OIB subroutine uses the unpack_ocn_ice_bdry subroutine from the ice_model module. The routine is changed to public and the use statement is added to the CIOD.

This PR will change answers for the coupled_AM2_LM3_SIS2/Intersperse_ice_1deg simulation, but should not change the solutions of any setup that does not use the interspersed option of the combined ice-ocean driver. The coupled_AM2_LM3_SIS2/Intersperse_ice_1deg simulation was run with a slightly modified version of the Xanadu coupler. Note that answers will only change when DT_COUPLED_ICE_OCEAN_DYN is less than dt_cpld. The USE_INTERSPERSE_BUG flag can be used to recover the old version.

### DIFF
--- a/src/combined_ice_ocean_driver.F90
+++ b/src/combined_ice_ocean_driver.F90
@@ -9,23 +9,26 @@ module combined_ice_ocean_driver
 ! This module provides a common interface for jointly stepping SIS2 and MOM6, and
 ! will evolve as a platform for tightly integrating the ocean and sea ice models.
 
-use MOM_coupler_types, only : coupler_type_copy_data, coupler_type_data_override
-use MOM_coupler_types, only : coupler_type_send_data
-use MOM_cpu_clock,     only : cpu_clock_id, cpu_clock_begin, cpu_clock_end, CLOCK_COMPONENT
-use MOM_data_override, only : data_override
-use MOM_domains,       only : domain2D, same_domain
-use MOM_error_handler, only : MOM_error, FATAL, WARNING, callTree_enter, callTree_leave
-use MOM_file_parser,   only : param_file_type, open_param_file, close_param_file
-use MOM_file_parser,   only : read_param, get_param, log_param, log_version
-use MOM_io,            only : file_exists, close_file, slasher, ensembler
-use MOM_io,            only : open_namelist_file, check_nml_error
-use MOM_time_manager,  only : time_type, time_type_to_real, real_to_time_type
-use MOM_time_manager,  only : operator(+), operator(-), operator(>)
+use MOM_coupler_types,  only : coupler_type_copy_data, coupler_type_data_override
+use MOM_coupler_types,  only : coupler_type_send_data
+use MOM_cpu_clock,      only : cpu_clock_id, cpu_clock_begin, cpu_clock_end, CLOCK_COMPONENT
+use MOM_data_override,  only : data_override
+use MOM_domains,        only : domain2D, same_domain
+use MOM_error_handler,  only : MOM_error, FATAL, WARNING, callTree_enter, callTree_leave
+use MOM_file_parser,    only : param_file_type, open_param_file, close_param_file
+use MOM_file_parser,    only : read_param, get_param, log_param, log_version
+use MOM_io,             only : file_exists, close_file, slasher, ensembler
+use MOM_io,             only : open_namelist_file, check_nml_error
+use MOM_time_manager,   only : time_type, time_type_to_real, real_to_time_type
+use MOM_time_manager,   only : operator(+), operator(-), operator(>)
 
-use ice_model_mod,     only : ice_data_type, ice_model_end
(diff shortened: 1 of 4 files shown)


src/ice_model.F90
```diff
@@ -2588,7 +2594,11 @@ subroutine ice_model_init(Ice, Time_Init, Time, Time_step_fast, Time_step_slow,
   else
     Ice%xtype = REDIST
   endif
-
+  
+  if (fast_ice_PE .eqv. slow_ice_PE) then
```
Hallberg-NOAA, line 2598: Should the `.eqv.` here be an `.and.`?

src/combined_ice_ocean_driver.F90
```diff
@@ -332,6 +356,71 @@ subroutine direct_flux_ice_to_IOB(Time, Ice, IOB, do_thermo)
...
+    intent(inout) :: Ice            !< The publicly visible ice data type in the slow part
+                                    !! of which the ocean surface information is to be stored.                                            
+  logical :: used, do_therm, do_area_weighted_flux
+
+  call cpu_clock_begin(fluxOceanIceClock)
+
+  do_therm = .true. ; if (present(do_thermo)) do_therm = do_thermo                                              
+  do_area_weighted_flux = .false. !! Need to add option to account for area weighted fluxes                           
+                                              
+  if( ASSOCIATED(OIB%u)     )OIB%u = Ocean%u_surf
```
Hallberg-NOAA, line 383: There should be white space after the `if` and between the closing ")" and OIB here and in the following lines in this routine, to read something like `if ( ASSOCIATED(OIB%u) ) OIB%u = Ocean%u_surf` to follow the MOM6 (and SIS2) style guide.

src/combined_ice_ocean_driver.F90
```diff
@@ -41,13 +43,16 @@ module combined_ice_ocean_driver
   logical :: intersperse_ice_ocn !< If true, intersperse the ice and ocean thermodynamic and
                               !! dynamic updates.  This requires the update ocean (MOM6) interfaces
                               !! used with single_MOM_call=.false. The default is false.
+  logical :: use_intersperse_bug !< If true, use a bug in the intersperse option where the ocean
```
Hallberg-NOAA, line 46: I am not completely sure of this, but it looks to me like the bug was the omission of these calls, so when USE_INTERSPERSE_BUG is true, the bug is being fixed.  We should change the logic around so that the use_bug = .true. option is the one with the bug and the one where this is false fixes the bug, to avoid user confusion.

theresa-morrison: Changed to be the correct logic.

src/ice_model.F90
theresa-morrison: This has been changed to .and. in the updated version.

theresa-morrison:
A new commit has been added to address the comments from the review.

marshallward:
This PR makes changes to the SIS2 API which causes issues with the current coupler.  We will discuss how to address this.

theresa-morrison:
When testing, I found that an additional halo update is required which has been added in SIS_dyn_trans.F90

kshedstrom:
This code calls open on SIS_input twice, not closing the file between open calls. gfortran complains with a fatal error.

Also, sooner or later you will get scolded for trailing white space. I found it best to teach my editor to make them go away on saving a file.

theresa-morrison:
Thanks for testing this, Kate. I think I fixed the issue if you want to try it again and eliminated most of the trailing white space.

kshedstrom:
Yes, I approve this PR.

Hallberg-NOAA:
This updated PR has passed pipeline testing at https://gitlab.gfdl.noaa.gov/ogrp/SIS2/-/pipelines/20132 .

PR #195, opened 2023-05-13
Fix for MOM6 issue #364

kshedstrom:
Corrects call to FMS2 interface requiring the domain argument.

### DIFF
--- a/src/SIS_restart.F90
+++ b/src/SIS_restart.F90
@@ -309,7 +309,7 @@ function open_restart_units(filename, directory, domain, CS, IO_handles, file_pa
         if (fexists) then
           nf = nf + 1
           if (present(IO_handles)) &
-            call open_file(IO_handles(nf), trim(filepath), READONLY_FILE, &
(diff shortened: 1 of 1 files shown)


marshallward (approved):
Thanks for finding and adding that, @kshedstrom

marshallward:
Now that our CI is finally working, I tested was able to test this.  Although it works with FMS2 I/O, it is failing with the FMS1.  I am fairly certain the problem is that the legacy `open_file()` wrapper function is not as robust as I had thought, and that is causing some issues with IO layouts.

I think that this will work if the instances of `file_type` and `field` are flipped over to `MOM_infra_file` and `MOM_field`, since the `open()` function does some extra work to spoof the IO_domain, but making those changes will be a bit more work.

I wither either submit those changes to this PR or put them in as a new one.  Either way, I'll keep you updated @kshedstrom.

marshallward:
I've opened #196 which should address this issue.

PR #198, opened 2023-08-14
+Remove rescaling factors from restart files

Hallberg-NOAA:
Removed the code to account for unit rescaling within the SIS2 restart files. This rescaling within the restart files has not been used in the code since the conversion arguments were added to all necessary register_restart_field calls on about June 2, 2022, as a part of PR #175.  The model will still work with older restart files provided that they did not use dimensional rescaling, and even if they did, they can be converted to avoid rescaling with a short run with the older code that created them.  This commit eliminates the publicly visible register_unit_conversion_restarts and rescale_fast_to_slow_restart_fields subroutines, as well as the call to fix_restart_unit_scaling.

  Large sections of rescale_ice_state_restart_fields were eliminated, but this routine is still needed to handle older (SIS1) restart files, a few of which are still used for initializing tests of SIS2.

  These changes significantly simplify the code, and they lead to a handful of constants that are always 1 not being included in the SIS2 restart files. All answers are bitwise identical, but two publicly visible interfaces have been eliminated.

### DIFF
--- a/src/SIS_dyn_cgrid.F90
+++ b/src/SIS_dyn_cgrid.F90
@@ -2083,7 +2083,6 @@ subroutine SIS_C_dyn_read_alt_restarts(CS, G, US, Ice_restart, restart_dir)
   ! then discarded.
   real, allocatable, target, dimension(:,:) :: str_tmp
   type(MOM_domain_type),   pointer :: domain_tmp => NULL()
-  real :: stress_rescale
   logical :: read_values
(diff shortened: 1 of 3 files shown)


marshallward (approved):

marshallward:
Gaea regression: https://gitlab.gfdl.noaa.gov/ogrp/SIS2/-/pipelines/21084 :heavy_check_mark:

PR #206, opened 2024-02-27
Reorder update and Ocn-to-OIB calls

theresa-morrison:
These changes move the ocean to OIB call to the beginning of the interspersed loop rather than the end so that it does not need to be called before the loop starts.

The needed translate OSS to sOSS has also been added to the direct flux OPT to OIB routine. Where variables are copied from the ocean public type (OPT) to the OIB "(:,:)" has been added.


MJHarrison-GFDL:
I have reviewed and tested these changes and agree with these changes.

PR #207, opened 2024-03-19
 To allow NCAR's MOM6 2024-03-19 #1621 to work

kshedstrom:
NCAR changed the interface to MOM_domains_init. We need to change accordingly.


kshedstrom:
I guess we won't be needing this.

PR #208, opened 2024-04-30
 Add diagnostics for true east/north velocities

kshedstrom:
This adds diagnostics for true north and east velocities to be output, required for our BOEM contract.


Hallberg-NOAA (approved):
I think that this would be a valuable new diagnostic addition even if it weren't required for the BOEM contract!

marshallward:
Gaea regression: https://gitlab.gfdl.noaa.gov/ogrp/mom6ci/SIS2/-/pipelines/23933 :heavy_check_mark: :yellow_circle:

PR #209, opened 2024-05-06
Initialization of point-particle bergs due to ice shelf calving

alex-huth:
-This PR is the SIS2 contribution for converting the MOM6 ice-shelf calving flux into new icebergs (this flux is routed through several model components as follows: MOM6 --> coupler --> SIS2 --> icebergs)
-The ice-shelf calving flux is added to FIA%calving through the new subroutine unpack_ocean_ice_boundary_calved_shelf_bergs; this is the ice-shelf equivalent to the existing subroutine
 unpack_land_ice_boundary, which adds the frozen discharge from land to  FIA%calving.
-Both styles of calving can occur in the same run (e.g. frozen discharge from Greenland and ice-shelf calving from Antarctica), as long as both calving fluxes are not non-zero in the same cell.

See related changes in:
Coupler: https://github.com/alex-huth/FMScoupler/commit/3e80bbc
MOM6: https://github.com/NOAA-GFDL/MOM6/pull/622

### DIFF
--- a/src/ice_boundary_types.F90
+++ b/src/ice_boundary_types.F90
@@ -35,8 +35,11 @@ type ocean_ice_boundary_type
     t      => NULL(), &  !< The ocean's surface temperature [Kelvin].
     s      => NULL(), &  !< The ocean's surface salinity [gSalt kg-1].
     frazil => NULL(), &  !< The frazil heat rejected by the ocean [J m-2].
-    sea_level => NULL()  !< The sea level after adjustment for any surface
-                         !! pressure that the ocean allows to be expressed [m].
+    sea_level => NULL(), & !< The sea level after adjustment for any surface
+                           !! pressure that the ocean allows to be expressed [m].
+    calving => NULL(), &   !< The mass flux per unit area of the ice shelf to convert to
+                           !!bergs [RZ_T ~> kg m-2 s-1].
+    calving_hflx => NULL() !< Calving heat flux [Q R Z T-1 ~> W m-2].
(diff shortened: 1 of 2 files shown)


jkrasting:
This PR requires Icepack to advance beyond the current v1.3.3 that is compatible with SIS @ dev/gfdl.  How do we reconcile this with @kshedstrom's PR (#202) that seems to be problematic.

@alex-huth -- would it be better to refactor this to work with Icepack v1.3.3 ?

(I'm a bit out of my realm here!  Hoping other with more expertise can weigh in.)

When trying to compile with Icepack v1.3.3:
```
SIS2/src/ice_ridge.F90(421): error #6784: The number of actual arguments cannot be greater than the number of dummy arguments.   [ICEPACK_STEP_RIDGE]
      call icepack_step_ridge (dt_sec,       ndtd,          &
-----------^
compilation aborted for SIS2/src/ice_ridge.F90 (code 1)
```

jkrasting:
Disregard my comment earlier.  If only this patch is merged into `dev/gfdl` there should be no issues.  The interfaces in SIS2 will still be consistent with Icepack v1.3.3.

marshallward (approved):
This looks very straightforward.  (I'm assuming this new function will be used in iOM4 or some other config.)

marshallward:
Gaea regression: https://gitlab.gfdl.noaa.gov/ogrp/mom6ci/SIS2/-/pipelines/23968 :heavy_check_mark:

PR #210, opened 2024-05-16
Add diag send complete calls

uramirez8707:
Adds a `SIS_diag_send_complete` subroutine which calls `diag_send_complete_infra` from MOM6. These are needed for the new diag manager but are backwards compatible with the old. 

This is called after every:
- `EVP_steps` in `SIS_C_dynamics`
-  `ndyn_steps` in `SIS_dynamics_trans`,  `SIS_merged_dyn_cont`, and `slab_ice_dyn_trans`
- `nsteps` in `ice_cat_transport`

It is also called before `disable_SIS_averaging` calls.

Unfortunately, this PR requires https://github.com/NOAA-GFDL/MOM6/pull/637

### DIFF
--- a/src/SIS_transport.F90
+++ b/src/SIS_transport.F90
@@ -15,6 +15,7 @@ use SIS_continuity,    only : continuity=>ice_continuity, SIS_continuity_CS
 use SIS_continuity,    only : summed_continuity, proportionate_continuity
 use SIS_diag_mediator, only : post_SIS_data, query_SIS_averaging_enabled, SIS_diag_ctrl
 use SIS_diag_mediator, only : register_diag_field=>register_SIS_diag_field, time_type
+use SIS_diag_mediator, only : SIS_diag_send_complete
 use SIS_framework,     only : safe_alloc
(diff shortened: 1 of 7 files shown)


marshallward (dismissed):
This looks correct and necessary to support the updated diag manager.

src/SIS_transport.F90
```diff
@@ -218,6 +219,7 @@ subroutine ice_cat_transport(CAS, TrReg, dt_slow, nsteps, G, US, IG, CS, uc, vc,
       write(mesg,'(i4)') n
       call check_SIS_tracer_bounds(TrReg, G, IG, "After advect_SIS_tracers "//trim(mesg))
     endif
+  call SIS_diag_send_complete()
```
marshallward: can you fix this indent?

marshallward:
Gaea regression: https://gitlab.gfdl.noaa.gov/ogrp/mom6ci/SIS2/-/pipelines/27355 :heavy_check_mark:

marshallward (approved):

PR #212, opened 2024-06-12
Add/modify sea ice diagnostics sivol, simass_n, siitdthick, sivol_CMOR

MitchBushuk:
Adds and modifies a number of sea ice diagnostics, namely:

-`sivol` is the equivalent thickness of sea ice (i.e. the grid-cell mean thickness). The prior `sivol` diagnostic was incorrect. It was equal to `sithick`, which is the floe thickness (i.e. the thickness of ice averaged over the ice-covered portion of the grid cell). This change removes the old `sivol` diagnostic, and renames the `sivol_CMOR` (which had the correct implementation of `sivol`) diagnostic to `sivol`. See Notz et al. (2016) for the SIMIP definitions of these diagnostics.
-The prior `simass_n` diagnostic is a confusing name, since it is actually the sea ice thickness in each category, not the mass in each category. This diagnostic has been modified such that `simass_n` is now the ice mass in each category. This now has the property that `sum_n simass_n = simass`.
-A new diagnostic `siitdthick` has been added, which is the ice thickness in each category, following the SIMIP variable definition.

### DIFF
--- a/src/SIS_ice_diags.F90
+++ b/src/SIS_ice_diags.F90
@@ -51,9 +51,9 @@ type, public :: ice_state_diags_type ; private
   integer :: id_cn = -1, id_hi = -1, id_hp = -1, id_hs = -1, id_tsn = -1, id_ext = -1
   integer :: id_t_iceav = -1, id_s_iceav = -1, id_e2m = -1, id_rdgf = -1, id_rdg_h = -1
 
-  integer :: id_simass = -1, id_simassn = -1, id_sisnmass = -1, id_sivol = -1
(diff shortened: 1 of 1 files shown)


theresa-cordero (dismissed):
This PR seems clear and it would help to have clearer variable names. It also seems to me that all old diagnostics will be recoverable although they may have different names.

adcroft (approved):
Passed with diagnostics changes at https://gitlab.gfdl.noaa.gov/ogrp/mom6ci/SIS2/-/pipelines/28768

Issue #213, opened 2024-08-12
Separate sea ice melt and direct precipitation to ocean when handling ice-ocean boundary mass fluxes


hdrake:
As commented by @StephenGriffies, there is a placeholder for a `seaice_melt` diagnostic in MOM6 but it currently does not get overwritten by SIS2, which bundles liquid precipitation together with sea ice melt here:

https://github.com/NOAA-GFDL/SIS2/blob/2c49005d1d7132776fff9b501cabe6ae85ef5f57/src/SIS_slow_thermo.F90#L1372-L1376

Because the ocean surface flux diagnostics get set by SIS2, this means that both liquid precipitation and sea ice melt/formation are included in MOM6's `prlq` ocean diagnostic, which really complicated process-based understanding of ocean mass budgets and salinity budgets.

It would be nice if SIS2 created a new `seaice_melt` variable that could be used to overwrite `IOB%seaice_melt` variable along with the others surface mass fluxes here:

https://github.com/NOAA-GFDL/SIS2/blob/2c49005d1d7132776fff9b501cabe6ae85ef5f57/src/combined_ice_ocean_driver.F90#L329-L332

hdrake:
The following shows that the piping is all there for these fluxes to propagate through to a new `seaice_melt` diagnostic in MOM6.

In MOM6, the `seaice_melt` array gets allocated here in the `ice_ocean_boundary`:

https://github.com/mom-ocean/MOM6/blob/aacb90989668f855bacc870637a1bb7abd8417c1/config_src/drivers/nuopc_cap/mom_cap.F90#L687

and filled with zeros here:

https://github.com/mom-ocean/MOM6/blob/aacb90989668f855bacc870637a1bb7abd8417c1/config_src/drivers/nuopc_cap/mom_cap.F90#L707

It does not seem to get updated anywhere else, and then is passed on to the surface fluxes type here:

https://github.com/mom-ocean/MOM6/blob/aacb90989668f855bacc870637a1bb7abd8417c1/config_src/drivers/nuopc_cap/mom_surface_forcing_nuopc.F90#L494-L496

and finally the diagnostic is registered here:

https://github.com/MOM6/MOM6/blob/ced41b5183920e0993f135feda0496a07a65981a/src/core/MOM_forcing_type.F90#L1313-L1320

and the data posted here:

[...]

PR #214, opened 2024-08-15
Separating sea ice melt/formation from liquid precipitation field

hdrake:
For unjustified legacy reasons, these have been bundled together in the past. They are not only bundled together in SIS2 variables, but also in the ice-ocean fluxes passed to the coupler and hence to MOM6.

This commit attempts to address this Issue: https://github.com/NOAA-GFDL/SIS2/issues/213

However, I currently am getting a SEGFAULT error stating that `iobt%seaice_melt` has not been allocated memory when I try to do a checksum on it.

(Edit: See compaion PRs https://github.com/NOAA-GFDL/MOM6/pull/710 and https://github.com/NOAA-GFDL/FMScoupler/pull/145)


hdrake:
This is the line that is currently giving me problems when running:

https://github.com/NOAA-GFDL/MOM6/blob/b3e65c63533bf17ae3dd70243f8827fa277ff386/config_src/drivers/FMS_cap/MOM_surface_forcing_gfdl.F90#L1799

If I remove it, the model runs (I am using the `Baltic_OM4_05` configuration for testing, with a `diag_table` that includes `fsitherm` from the `ocean_model`) but then the `fsitherm` diagnostic stays full of zero and is not being currently overwritten by the coupling with SIS2.

hdrake:
@theresa-morrison pointed out that that I forgot to allocate the new `seaice_melt` field in the coupler!

After doing so [in this PR](https://github.com/NOAA-GFDL/FMScoupler/pull/145), my Baltic sea test now works!

Screenshot 2024-08-16 at 4 58 55 PM

src/ice_model.F90
```diff
@@ -659,7 +661,7 @@ subroutine set_ocean_top_fluxes(Ice, IST, IOF, FIA, OSS, G, US, IG, sCS)
   if (allocated(IOF%melt_nudge)) then
     do j=jsc,jec ; do i=isc,iec
       i2 = i+i_off ; j2 = j+j_off! Use these to correct for indexing differences.
-      Ice%lprec(i2,j2) = Ice%lprec(i2,j2) + US%RZ_T_to_kg_m2s*IOF%melt_nudge(i,j)
+      Ice%seaice_melt(i2,j2) = Ice%seaice_melt(i2,j2) + US%RZ_T_to_kg_m2s*IOF%melt_nudge(i,j)
```
hdrake, line 664: @theresa-morrison, I realized I don't really know what `melt_nudge` is for, but I assumed this correction should be moved from `lprec` to the new `seaice_melt` diagnostic?

theresa-cordero: I would say so. It looks like melt associated with artificially nudging the sea ice concentration to match observations. See ~line 713 of SIS_slow_thermo. It should be zero in the Baltic test cases.

src/SIS_slow_thermo.F90
```diff
@@ -857,6 +854,11 @@ subroutine SIS2_thermodynamics(IST, dt_slow, CS, OSS, FIA, IOF, G, US, IG)
     IOF%lprec_ocn_top(i,j) = IOF%lprec_ocn_top(i,j) + IST%part_size(i,j,k) * FIA%lprec_top(i,j,k)
   enddo ; enddo ; enddo
 
+  ! Reset sea ice melt rate to zero
+  do j=jsc,jec ; do i=isc,iec
+    IOF%seaice_melt(i,j) = 0.
```
theresa-cordero: does this need to be a loop or would `IOF%seaice_melt(:,:)=0.` work just as well?

theresa-cordero:
Is there any reason we would want to be able to save the old method for calculating net_melt = melt+ lprec? Or is prlq (new) + fsitherm = prlq (old) sufficient for recreating old results if needed?

src/SIS_slow_thermo.F90
hdrake: I think you're probably right. I can test it.

hdrake:
> Is there any reason we would want to be able to save the old method for calculating net_melt = melt+ lprec? Or is prlq (new) + fsitherm = prlq (old) sufficient for recreating old results if needed?

I personally think this would be undesirable, as it just propagates something that was already mislabelled and could lead to further misinterpretation.

Hallberg-NOAA:
We always retain the ability to use the old calculations if there is any chance at all that anyone is using them, no matter how ugly the code was.  We can wrap in a runtime BUG flag or an ANSWER_DATE flag, but we do not ever change other people's answers without warning and explicit consent.  Keeping the new method as the only option is only sufficient if it can be configured to be bitwise identical to the old one!

hdrake:
Thanks for clarifying @Hallberg-NOAA. I do think answers will not change in the sense that the sum of the two existing diagnostics, `prlq + fsitherm`, will be bitwise identical. But for `prlq` and `fsitherm` individually the answers will definitely change.

Is this fine or should I add a runtime BUG flag? I assume I can follow the implementation for something like this:
https://github.com/NOAA-GFDL/SIS2/blob/2c49005d1d7132776fff9b501cabe6ae85ef5f57/src/ice_model.F90#L214-L223

src/SIS_slow_thermo.F90
hdrake: Fixed in https://github.com/NOAA-GFDL/SIS2/pull/214/commits/49a2829d4f9f6f1024b119eb4a35c1398e935f14

src/SIS_slow_thermo.F90
```diff
@@ -1361,20 +1361,14 @@ subroutine SIS2_thermodynamics(IST, dt_slow, CS, OSS, FIA, IOF, G, US, IG)
                                     scale=US%RZ_T_to_kg_m2s*Idt_slow)
   if (CS%id_qflim>0) call post_data(CS%id_qflim, qflx_lim_ice, CS%diag)
   if (CS%id_qfres>0) call post_data(CS%id_qfres, qflx_res_ice, CS%diag)
-  if (CS%id_net_melt>0) call post_data(CS%id_net_melt, net_melt, CS%diag)
-  if (CS%id_CMOR_melt>0) call post_data(CS%id_CMOR_melt, net_melt, CS%diag)
+  if (CS%id_net_melt>0) call post_data(CS%id_net_melt, IOF%seaice_melt, CS%diag)
+  if (CS%id_CMOR_melt>0) call post_data(CS%id_CMOR_melt, IOF%seaice_melt, CS%diag)
```
theresa-cordero, line 1365: to reproduce old answers, I think this should be okay, since `IOF%seaice_melt` and `net_melt` are the same. Right?

src/SIS_slow_thermo.F90
```diff
@@ -1361,20 +1361,14 @@ subroutine SIS2_thermodynamics(IST, dt_slow, CS, OSS, FIA, IOF, G, US, IG)
...
 
   if (coupler_type_initialized(IOF%tr_flux_ocn_top)) &
     call coupler_type_send_data(IOF%tr_flux_ocn_top, CS%Time)
 
   call disable_SIS_averaging(CS%diag)
 
-  ! Combine the liquid precipitation with the net melt of ice and snow for
```
theresa-cordero, line 1372: I think a bug fix flag here will work to get the old diagnostics, but it would be ideal (in my opinion) if there was a way to save `sea ice melt` and `sea ice melt + lprec` from the same simulation.

src/SIS_sum_output.F90
```diff
@@ -759,6 +759,7 @@ subroutine accumulate_bottom_input(IST, OSS, FIA, IOF, dt, G, US, IG, CS)
   do j=jsc,jec ; do i=isc,iec
     CS%water_in_col(i,j) = CS%water_in_col(i,j) - dt * &
            ( ((FIA%runoff(i,j) + FIA%calving(i,j)) + &
+               IOF%seaice_melt(i,j) + &
```
Hallberg-NOAA: Please add parentheses to specify the order with which `IOF%seaice_melt` is added to the other fields so that this diagnostic does not capriciously change with the level of optimization.
My preference would be for `IOF%seaice_melt` to be added first to `IOF%lprec_ocn_top` so that the answers more closely resemble what was done previously, but other options could be justified as long as the order of arithmetic is specified unambiguously.

hdrake: Thank you @Hallberg-NOAA; I've started to make the necessary changes wherever `fluxes%lprec` shows up.

I am not sure what to do here:
https://github.com/NOAA-GFDL/MOM6/blob/37d1385b2979a453c5b2c24d91bbea7c74c6a82e/src/core/MOM_forcing_type.F90#L2275

Is there any way to separately increment these averages for `fluxes%lprec` and `fluxes%seaice_melt` while preserving the overall order of operations?

hdrake: Similarly, in any `if associated(fluxes%lprec)` blocks, do I need to separately treat the two cases of `if associated(fluxes%seaice_melt)`?

theresa-cordero:
Henri, we are planning to merge MOM6 and SIS2 in the near future (hopefully by the end of the calendar year at the latest). Combining the two code bases would make merging both this PR and the related MOM6 PR much easier. Would you be willing to wait until the merge happens for these code changes to be accepted?

hdrake:
Oh, that's great. Yes, it would be much easier to reimplement these changes within the same code base.

I figured out how to separate these fields using the sea ice diagnostics offline in the meantime anyways.

PR #215, opened 2024-10-25
 This is the code update for calling new Icepack.

kshedstrom:
The "main" branch as of Oct 24, 2024 adds the optional arguments
 described here.

 It uses the public icepack_step_ridge instead of the old interface,
 which is now private. The code that was in Icepack's ridge_prep had
 to be moved to the SIS2 side. There are two new options in the call to
 icepack_step_ridge, telling it how much to do. To reproduce old answers,
 both options should be set to False. On the Icepack side, they recommend
 both being True, which will not only call the ridging code, but also put
 a limiter on some ice fields (no negative thicknesses, etc.) and call
 the rebin routine, which resets which ice is in which thickness category.

 All three versions run for me (three because rebin depends on calling
 the cleanup routine). I have promised some feedback if we later decide
 to just pick one forever.

### DIFF
--- a/src/ice_ridge.F90
+++ b/src/ice_ridge.F90
@@ -23,14 +23,14 @@ use MOM_error_handler, only : SIS_error=>MOM_error, FATAL, WARNING, SIS_mesg=>MO
 use MOM_file_parser,   only : get_param, log_param, read_param, log_version, param_file_type
 use MOM_unit_scaling,  only : unit_scale_type
 use SIS_hor_grid,      only : SIS_hor_grid_type
-use SIS_types,         only : ice_state_type, ist_chksum
+use SIS_types,         only : ice_state_type, ist_chksum, ocean_sfc_state_type
 use SIS_tracer_registry, only : SIS_tracer_registry_type, SIS_tracer_type, get_SIS_tracer_pointer
-use SIS2_ice_thm,      only : get_SIS2_thermo_coefs
+use SIS2_ice_thm,      only : get_SIS2_thermo_coefs, T_freeze
 use ice_grid,          only : ice_grid_type
 !Icepack modules
 use icepack_kinds
 use icepack_itd, only: icepack_init_itd, cleanup_itd
-use icepack_mechred, only: ridge_ice
+use icepack_mechred, only: icepack_step_ridge
 use icepack_warnings, only: icepack_warnings_flush, icepack_warnings_aborted, &
                             icepack_warnings_setabort
 use icepack_tracers, only: icepack_init_tracer_indices, icepack_init_tracer_sizes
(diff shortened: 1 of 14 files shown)


kshedstrom:
Icepack 1.5.0 has been released with this note:
NOTE: This version is not backwards compatible with earlier versions of Icepack due to changes in public interfaces. In addition, all answers are changed.

MJHarrison-GFDL:
Thanks for your efforts Kate! We will test the various options.



On Tuesday, November 19, 2024, Kate Hedstrom ***@***.***>
wrote:
> Icepack 1.5.0 has been released with this note:
> NOTE: This version is not backwards compatible with earlier versions of
Icepack due to changes in public interfaces. In addition, all answers are
changed.
>
> —
> Reply to this email directly, view it on GitHub, or unsubscribe.
> You are receiving this because you are subscribed to this thread.<
https://ci3.googleusercontent.com/meips/ADKq_NY6kGvG0z62AUX5sETqmueGGpwmBiQE9dk-X_ImHy1_1TSHhjU1tHb6y5BShp6ZwKK5-9ztmIKexNalsJ9p7HPUmZGbWA8EQW88bWbGtAsAc2a1mvfG1bykDO1XzJeuXvkc_ka8pb2gmyuWfsGjYewQzO3aVA8ZWzBDj8J5kSZvlJvtj535r7ht5ikFR6KVtzSaxuL9huYyFGsp_6yPN097kIQ3Sh_RExAdLD7LFYKgASTzh8eOTnY=s0-d-e1-ft#https://github.com/notifications/beacon/ABSPZ4FZCEOKVYOEQXCVNT32BO3TNA5CNFSM6AAAAABQT6NWHWWGG33NNVSW45C7OR4XAZNMJFZXG5LFINXW23LFNZ2KUY3PNVWWK3TUL5UWJTUUHLGOA.gif>Message
ID: ***@***.***>

-- 
Matthew Harrison (he/him)
USDOC/NOAA/GFDL/Oceans&Cryosphere

theresa-cordero:
I am I correctly understanding that these code changes would mean users have to use an updated version of  Icepack? And that the answers will change with the updated version of Icepack?

theresa-cordero:
@MitchBushuk 
Besides Kate in the regional Arctic domain, we thought that OM5 is the only other configuration using these Icepack options. Is that correct? Would it be okay for OM5 to change answers because of a change in the Icepack version?

kshedstrom:
This adds two flags:
> To reproduce old answers, both options should be set to False.

Yes, you have to update Icepack.

MitchBushuk:
Thanks @kshedstrom for your great work on this PR! Just to confirm, if the flags `docleanup` and `dorebin` are both set to False, this should reproduce old answers, correct? 

Also, if these flags are set to True, does this change the model's ability to use the `krdg_partic=1` and `krdg_redist=1` options? The earlier version would crash with tiny negative thickness values when these options were set, but I wonder if the new cleanup routine fixes this.

@theresa-cordero it sounds like Kate's flags allow us to reproduce old answers, so we should be all set from an OM5 perspective.

kshedstrom:
Yes, I can reproduce old answers with the flags set to False.

I too was hoping that we can try the new ridging with Icepack doing the cleanup, but no, it didn't work for me.

marshallward (dismissed):

kshedstrom:
I believe it is running now with the new ridging scheme.

src/ice_ridge.F90
```diff
@@ -292,231 +322,248 @@ subroutine ice_ridging(IST, G, IG, mca_ice, mca_snow, mca_pond, TrReg, CS, US, d
...
-
-      del_sh = sqrt(sh_Dd**2 + 0.25 * (sh_Dt**2 + &
-                   (0.25 * ((sh_Ds(I-1,J-1) + sh_Ds(I,J)) + &
-                            (sh_Ds(I-1,J) + sh_Ds(I,J-1))))**2 ) )*US%s_to_T ! H&D eqn 9
-      rdg_conv  = -min(sh_Dd,0.0)*US%s_to_T              ! energy dissipated by convergence ...
-      rdg_shear = 0.5*(del_sh-abs(sh_Dd))*US%s_to_T      ! ... and by shear
-
-      aice0 = IST%part_size(i,j,0)
-      if (aice0<0.) then
-         call SIS_error(WARNING,'aice0<0. before call to ridge ice.')
```
theresa-cordero, line 329: Could we keep this warning? It would be good to know if we have negative ocean area.

kshedstrom: We can keep it, though it is just a sign that ridging needs to happen.

theresa-cordero: oh, so it is "okay" to have negative ocean area? 
Weirdly, I have a run I am debugging right now where the part_size(0)<0 in the checksum but this aice0<0 warning is not showing up. 
Should I switch to this branch and ridging scheme for the ARC12?

kshedstrom: The model needs a way to keep the sum of ice part_size to one or less. This is through ridging either via Icepack or the alternative.

theresa-cordero (approved):
I approve these changes and was able to compile and run with a newer version of Icepack. 

Note that although this branch is titled icepack_1.4_mods, the code changes have been updated to work with the Icepack 1.5.1 release.

kshedstrom:
You are right, this needs Icepack tag 1.5.1.

marshallward:
Gaea regression: https://gitlab.gfdl.noaa.gov/ogrp/mom6ci/SIS2/-/pipelines/28651

But as we discussed, OM5 should do its own regression test before merging.

MitchBushuk:
I have successfully compiled OM5 with SIS2 PR-215 and Icepack 1.5.1 (hash b91f1dd73d6c8475e6d94611f592b8036623bf78), and confirmed that this does not change answers in OM5. This is expected behavior, as the standard OM5 setup does not have sea ice ridging turned on. From the OM5 standpoint, this PR is ready to be merged.

PR #216, opened 2025-03-13
SIS2 sea ice thickness and partial area relaxation

DmitryDukhovskoy:
Suggested changes include subroutines for implementing Newtonian relaxation of sea ice thickness and partial area (concentration) fields, which will relax the SIS2 predicted values toward the target values over a specified region with a prescribed relaxation rate.  The relaxation algorithm can serve as an alternative to the sea ice open boundary conditions in regional applications or when sea ice fields need to be nudged or corrected toward reference values within the domain. The main subroutines for initializing, processing, interpolating the target fields, and applying the relaxation are contained in the module SIS_sponge.F90. Necessary modifications have been made to several other SIS2 codes. To run the SIS2 code with relaxation, the user must turn SIS_SPONGE flag "on" (True), specify the relaxation parameters in SIS_input and provide an input file containing a 2D sea ice relaxation rate field, a 2D field of sea ice thicknesses (in meters), and the partial area. The relaxation code will then distribute ice thicknesses and partial areas by category.

### DIFF
--- a/src/SIS_slow_thermo.F90
+++ b/src/SIS_slow_thermo.F90
@@ -1559,14 +1559,8 @@ subroutine SIS_slow_thermo_init(Time, G, US, IG, param_file, diag, CS, tracer_fl
                  "If true, SIS sponges may be applied anywhere in the domain. "//&
                  "The exact location and properties of those sponges are "//&
                  "specified via SIS_SPONGE_CONFIG.", default=.false.)
-  if (use_isponge) then                              
-    call SIS_mesg("SIS_slow_thermo: returned ICE SPONGE flag: True")
-    call SIS_mesg("SIS_slow_thermo: calling initialize_icerelax_file")
+  if (use_isponge) &
     call initialize_icerelax_file(param_file, G, IG, ispCS, US, sIST, Time)
-  else
-    call SIS_mesg("SIS_slow_thermo: returned ICE SPONGE flag: False")
-  endif
-
 
   CS%id_lsrc = register_diag_field('ice_model','LSRC', diag%axesT1, Time, &
                'frozen water local source', 'kg/(m^2*yr)', missing_value=missing)
--- a/src/SIS_slow_thermo.F90
+++ b/src/SIS_slow_thermo.F90
@@ -59,8 +59,7 @@ use SIS2_ice_thm,      only : SIS2_ice_thm_CS, SIS2_ice_thm_init, SIS2_ice_thm_e
 use SIS2_ice_thm,      only : ice_resize_SIS2, add_frazil_SIS2, rebalance_ice_layers
 use SIS2_ice_thm,      only : get_SIS2_thermo_coefs, enthalpy_liquid_freeze
 use SIS2_ice_thm,      only : enth_from_TS, Temp_from_En_S, enthalpy_liquid, calculate_T_freeze
-use SIS_sponge,        only : initialize_icerelax_file, set_up_isponge_field, apply_isponge, isponge_CS
-use SIS_sponge,        only : global_to_local_ij
+use SIS_sponge,        only : initialize_icerelax_file, apply_isponge, isponge_CS
 
 implicit none ; private
 
--- a/src/SIS_slow_thermo.F90
+++ b/src/SIS_slow_thermo.F90
@@ -59,6 +59,8 @@ use SIS2_ice_thm,      only : SIS2_ice_thm_CS, SIS2_ice_thm_init, SIS2_ice_thm_e
 use SIS2_ice_thm,      only : ice_resize_SIS2, add_frazil_SIS2, rebalance_ice_layers
 use SIS2_ice_thm,      only : get_SIS2_thermo_coefs, enthalpy_liquid_freeze
 use SIS2_ice_thm,      only : enth_from_TS, Temp_from_En_S, enthalpy_liquid, calculate_T_freeze
+use SIS_sponge,        only : initialize_icerelax_file, set_up_isponge_field, apply_isponge, isponge_CS
+use SIS_sponge,        only : global_to_local_ij
 
 implicit none ; private
 
@@ -293,7 +295,7 @@ end subroutine post_flux_diagnostics
 
 !~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~!
 !> slow_thermodynamics takes care of slow ice thermodynamics and mass changes
-subroutine slow_thermodynamics(IST, dt_slow, CS, OSS, FIA, XSF, IOF, G, US, IG)
+subroutine slow_thermodynamics(IST, dt_slow, CS, OSS, FIA, XSF, IOF, G, US, IG, ispCS)
 
   type(ice_state_type),       intent(inout) :: IST !< A type describing the state of the sea ice
   real,                       intent(in)    :: dt_slow !< The thermodynamic step [T ~> s].
@@ -310,6 +312,7 @@ subroutine slow_thermodynamics(IST, dt_slow, CS, OSS, FIA, XSF, IOF, G, US, IG)
   type(SIS_hor_grid_type),    intent(inout) :: G   !< The horizontal grid type
   type(unit_scale_type),      intent(in)    :: US  !< A structure with unit conversion factors
   type(ice_grid_type),        intent(inout) :: IG  !< The sea-ice specific grid type
+  type(isponge_CS),           pointer       :: ispCS !< A control structure for ice relaxation fields
 
   ! Local variables
   real, dimension(SZI_(G),SZJ_(G))   :: &
@@ -423,6 +426,9 @@ subroutine slow_thermodynamics(IST, dt_slow, CS, OSS, FIA, XSF, IOF, G, US, IG)
     IOF%Enth_Mass_in_atm(i,j) = 0.0 ; IOF%Enth_Mass_out_atm(i,j) = 0.0
     IOF%Enth_Mass_in_ocn(i,j) = 0.0 ; IOF%Enth_Mass_out_ocn(i,j) = 0.0
   enddo ; enddo
+
+  if (ispCS%use_isponge) &
+    call apply_isponge(dt_slow, ispCS, G, IG, IST, US, OSS, CS%Time)
 
   ! The thermodynamics routines return updated values of the ice and snow
   ! masses-per-unit area and enthalpies.
@@ -1383,7 +1389,8 @@ end subroutine SIS2_thermodynamics
 !~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~!
 !> SIS_slow_thermo_init - initializes the parameters and diagnostics associated
 !!    with the SIS_slow_thermo module.
-subroutine SIS_slow_thermo_init(Time, G, US, IG, param_file, diag, CS, tracer_flow_CSp)
+subroutine SIS_slow_thermo_init(Time, G, US, IG, param_file, diag, CS, tracer_flow_CSp, &
+                                ispCS, sIST)
   type(time_type),     target, intent(in)    :: Time !< The sea-ice model's clock,
                                                      !! set with the current model.
   type(SIS_hor_grid_type),     intent(in)    :: G    !< The horizontal grid structure
@@ -1396,11 +1403,14 @@ subroutine SIS_slow_thermo_init(Time, G, US, IG, param_file, diag, CS, tracer_fl
   type(SIS_tracer_flow_control_CS), &
                                pointer       :: tracer_flow_CSp !< A structure that is used to
                                                      !! orchestrate the calling ice tracer packages
+  type(isponge_CS),            pointer       :: ispCS !< A pointer to the structure for ice sponge/relax fields
+  type(ice_state_type),        pointer       :: sIST  !< A pointer to sea ice state fields
 
 ! This include declares and sets the variable "version".
 #include "version_variable.h"
   character(len=40) :: mdl = "SIS_slow_thermo" ! This module's name.
   logical           :: debug
+  logical           :: use_isponge
   real               :: transmute_scale ! A scaling factor to use when reading the transmutation rate.
   character(len=64)  :: transmute_var   ! Transmutation rate variable name in file
   character(len=200) :: filename, transmute_file, inputdir ! Strings for file/path
@@ -1545,6 +1555,18 @@ subroutine SIS_slow_thermo_init(Time, G, US, IG, param_file, diag, CS, tracer_fl
                  "sensible, and issue warnings if they are not.  This "//&
                  "does not change answers, but can increase model run time.", &
                  default=.true.)
+  call get_param(param_file, mdl, "SIS_SPONGE", use_isponge, &
+                 "If true, SIS sponges may be applied anywhere in the domain. "//&
+                 "The exact location and properties of those sponges are "//&
+                 "specified via SIS_SPONGE_CONFIG.", default=.false.)
+  if (use_isponge) then                              
+    call SIS_mesg("SIS_slow_thermo: returned ICE SPONGE flag: True")
+    call SIS_mesg("SIS_slow_thermo: calling initialize_icerelax_file")
+    call initialize_icerelax_file(param_file, G, IG, ispCS, US, sIST, Time)
+  else
+    call SIS_mesg("SIS_slow_thermo: returned ICE SPONGE flag: False")
+  endif
+
 
   CS%id_lsrc = register_diag_field('ice_model','LSRC', diag%axesT1, Time, &
                'frozen water local source', 'kg/(m^2*yr)', missing_value=missing)
--- a/src/SIS_slow_thermo.F90
+++ b/src/SIS_slow_thermo.F90
@@ -1558,8 +1558,12 @@ subroutine SIS_slow_thermo_init(Time, G, US, IG, param_file, diag, CS, tracer_fl
                  "If true, SIS sponges may be applied anywhere in the domain. "//&
                  "The exact location and properties of those sponges are "//&
                  "specified via SIS_SPONGE_CONFIG.", default=.false.)
-  if (use_isponge) &
+  if (use_isponge) then
     call initialize_icerelax_file(param_file, G, IG, ispCS, US, sIST, Time)
+  else
+    allocate(ispCS)
+    ispCS%use_isponge = .false.
+  endif
 
   CS%id_lsrc = register_diag_field('ice_model','LSRC', diag%axesT1, Time, &
                'frozen water local source', 'kg/(m^2*yr)', missing_value=missing)
--- a/src/SIS_sponge.F90
+++ b/src/SIS_sponge.F90
@@ -154,8 +154,8 @@ subroutine initialize_icerelax_file(param_file, G, IG, CS, US, IST, Time)
                  "The name of the ice thickness variable in "//&
                  "ISPONGE_STATE_FILE.", default="ithkn")
   call get_param(param_file, mdl, "ISPONGE_IAREA_VAR", iarea_var, &
-                 "The name of the ice partial area/concentration variable in "//&
-                 "ISPONGE_STATE_FILE.", default="ithkn")
+                 "The name of the ice partial area variable in "//&
+                 "ISPONGE_STATE_FILE.", default="iarea")
   call get_param(param_file, mdl, "ISPONGE_RLXRATE_VAR", rlxrate_var, &
                  "The name of the relaxation rate variable in "//&
                  "ISPONGE_RELAX_FILE.", default="relax_rate")
--- a/src/SIS_sponge.F90
+++ b/src/SIS_sponge.F90
@@ -74,7 +74,6 @@ end type f3d
 !> This control structure holds memory and parameters for the SIS_sponge module
 type, public :: isponge_CS ; private
   logical, public :: use_isponge = .false.  !< If true, ice tracer fields may be relaxed somewhere in the domain
-  integer, public :: itest, jtest    !< Test point where diagnostics are printed out for checking the relaxation
   integer         :: num_col         !< The number of relaxation points within the computational domain.
   integer, public :: fldno = 0       !< The number of fields which have already been
                                      !! registered by calls to set_up_isponge_field
@@ -119,7 +118,6 @@ subroutine initialize_icerelax_file(param_file, G, IG, CS, US, IST, Time)
   integer :: i, j, k, is, ie, js, je, ncat
   integer :: isd, ied, jsd, jed
   integer :: isc, iec, jsc, jec
-  integer :: itestG, jtestG, itest, jtest  ! Test point indices global and local
   integer :: year     !< The current model year
   integer :: day      !< The current model year-day
   integer :: second   !< The second of the day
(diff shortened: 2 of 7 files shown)


src/SIS_sponge.F90
```diff
@@ -0,0 +1,989 @@
...
+  filename = trim(inputdir)//trim(relaxrate_file)
+  call log_param(param_file, mdl, "INPUTDIR/ISPONGE_RELAX_FILE", filename)
+  if (.not.file_exists(filename, G%Domain)) &
+    call SIS_error(FATAL, " initialize_icerelax_file: Unable to open "//trim(filename))
+
+  call MOM_read_data(filename, rlxrate_var, Irelax(:,:), G%Domain, scale=US%s_to_T)
+  max_rlxrate =  maxval(Irelax*US%T_to_s)
+  call max_across_PEs(max_rlxrate)
+  call get_date(Time, year, mon, day, hr, minute, second, itick)
+  write(mesg,'("SIS Time:",i6,2("/",i2.2),1x,3(":",i2.2),"; max(Irelax)=",D13.4," s-1")') &
```
theresa-cordero: I like that you have added this sanity check, but I think it would be better to include logic to only write these extra messages if `DEBUG= True`.

src/SIS_sponge.F90
```diff
@@ -0,0 +1,989 @@
...
+
+! Set default, read and log parameters
+! get_param (procedure --> get_param_logical) - checks if variable is set true in the param_file
+  call log_version(param_file, mdl, version)
+  call get_param(param_file, mdl, "SIS_SPONGE", use_isponge, &
+                 "If true, sponges may be applied anywhere in the domain. "//&
+                 "The exact location and properties of those sponges are "//&
+                 "specified from MOM_initialization.F90.", default=.false.)
+
+  if (use_isponge) then
```
theresa-cordero: I think these messages are redundant because the SIS parameter file will have SIS_SPONGE=True or False

theresa-cordero:
Thanks for this great addition to SIS2! 

I have taken an initial look at the code and made a couple minor suggests. I also asked @Hallberg-NOAA and @marshallward to take a closer look at the code, especially the IO.

src/SIS_sponge.F90
DmitryDukhovskoy: Sure, this can be deleted. This is a leftover from my development version of the code. I wanted to have all information in the log file.

src/SIS_sponge.F90
DmitryDukhovskoy: I'm not sure about this. I liked to have this option separate from the overall debug to be able to track the relaxation without printing out anything else. Having this information was very helpful to check the values with the target fields. Using "DEBUG" would complicate things for the users. Right now, it is very easy to add/not this information by simply specifying the test point in the SIS_input. Adding debug will require the user to remember to turn this flag on in addition to adding the test point, and if the test point is not needed, then the user will have to remember to turn the debug flag off. Besides, having debug on, will add lots of unneeded information if the only purpose is to check ice relaxation.

DmitryDukhovskoy:
Theresa, thank you for looking into the code. I'll wait for comments from others to make all final changes.

kshedstrom:
How about a DEBUG_SPONGE option?

DmitryDukhovskoy:
> How about a DEBUG_SPONGE option?

Sure, this is an option. Still the user will need to specify the test point where he/she wants the diagnostics to be printed out, right?  Does this mean that if the test points are provided but DEBUG_SPONGE=False, then no diagnostics are printed out (the test points ignored), similarly if the DEBUG_SPONGE=True but no test point is specified - no diagnostics. Right?

kshedstrom:
If you write it that way, sure.

src/SIS_sponge.F90
marshallward: These messages appear to always be printed, is that right?  I see a lot of `SIS_mesg()` calls in the function.

We generally try to minimize diagnostic output like this during runtime.  If every module were to report its status, then warnings and other serious messages could be easily missed.

One solution is to add a verbosity level to `SIS_mesg(message, verb)`.  The default value is 2, so anything higher will be removed.  A higher value, say 5, is sometimes used to print diagnostic info.  `DEBUG` corresponds to 9, which prints everything.

Hopefully this avoids any questions about complex juggling of multiple `DEBUG*` flags.

You might also want to wrap the `max_across_PEs` around a verbosity flag.  It's always a good idea to avoid collectives if you can manage it.

(Unrelated point, but `.lt.` and `.gt.` should also be replace with `<` and `>` when you get a chance.)

DmitryDukhovskoy: Thanks Marshall, I agree.  With the test point information, would this be ok if I leave it as is? It is easy to turn on/off this diagnostics in SIS_input by specifying or not the test point indices. Otherwise it would require the user to specify both the test point and verbosity level. Or how would you do this?

marshallward: Using `ISPONGE_[IJ]TEST` to control output seems ok to me, although I don't know if there's any precedent for this sort of output.

The closest I can think of is truncation reporting in (say) `MOM_vert_friction.F90` or `MOM_PointAccel.F90` which logs invalid states to a file, rather than monitoring a designated point.

I'd say leave it for now but let's see what @Hallberg-NOAA thinks when he returns next week.

DmitryDukhovskoy: Sounds good, thanks! In the latest commit, I think I addressed all comments including "max_across_PEs".

src/SIS_slow_thermo.F90
```diff
@@ -59,6 +59,8 @@ module SIS_slow_thermo
 use SIS2_ice_thm,      only : ice_resize_SIS2, add_frazil_SIS2, rebalance_ice_layers
 use SIS2_ice_thm,      only : get_SIS2_thermo_coefs, enthalpy_liquid_freeze
 use SIS2_ice_thm,      only : enth_from_TS, Temp_from_En_S, enthalpy_liquid, calculate_T_freeze
+use SIS_sponge,        only : initialize_icerelax_file, set_up_isponge_field, apply_isponge, isponge_CS
+use SIS_sponge,        only : global_to_local_ij
```
Hallberg-NOAA, line 64: `global_to_local_ij()` and `set_up_isponge_feild` are not used in SIS_slow_thermo.F90, and hence should be remove from these `use ..., only` statements.

src/SIS_sponge.F90
```diff
@@ -0,0 +1,991 @@
...
+  integer :: isd, ied, jsd, jed
+  integer :: isc, iec, jsc, jec
+  integer :: itestG, jtestG, itest, jtest  ! Test point indices global and local
+  integer :: year     !< The current model year
+  integer :: day      !< The current model year-day
+  integer :: second   !< The second of the day
+  integer :: mon, hr, minute, itick
+  integer :: start_of_day, num_days
+  integer :: verbosity   !< MOM verbosity level
+  real :: max_rlxrate, rho_ice
```
Hallberg-NOAA: Please add comments describing the real variables `max_rlxrate` and `rho_ice` along with their units.

src/SIS_sponge.F90
```diff
@@ -0,0 +1,991 @@
...
+use MOM_io,            only : file_exists, MOM_read_data, slasher
+use MOM_io,            only : axis_info
+use MOM_interpolate,   only : init_external_field, get_external_field_info, time_interp_external_init
+use MOM_interpolate,   only : time_interp_external
+use MOM_interpolate,   only : external_field
+
+use SIS_diag_mediator, only : post_SIS_data, post_data=>post_SIS_data
+use SIS_diag_mediator, only : register_diag_field=>register_SIS_diag_field
+use SIS_hor_grid,      only : SIS_hor_grid_type
+use SIS_sum_output,    only : SIS_sum_out_CS, write_ice_statistics! , SIS_sum_output_init
```
Hallberg-NOAA: I don't believe that `write_ice_statistics()` or `SIS_sum_out_CS` are used in the SIS_sponge.F90 module.

src/SIS_sponge.F90
```diff
@@ -0,0 +1,991 @@
...
+  ! Read in relaxation rate, s-1, for ice thickness and partial area
+  filename = trim(inputdir)//trim(relaxrate_file)
+  call log_param(param_file, mdl, "INPUTDIR/ISPONGE_RELAX_FILE", filename)
+  if (.not.file_exists(filename, G%Domain)) &
+    call SIS_error(FATAL, " initialize_icerelax_file: Unable to open "//trim(filename))
+
+  call MOM_read_data(filename, rlxrate_var, Irelax(:,:), G%Domain, scale=US%s_to_T)
+  max_rlxrate =  maxval(Irelax*US%T_to_s)
+  call get_date(Time, year, mon, day, hr, minute, second, itick)
+  call get_time(Time, start_of_day, num_days)
```
Hallberg-NOAA: I do not believe that `start_of_day` or `num_days` are used.

src/SIS_sponge.F90
```diff
@@ -0,0 +1,991 @@
...
+                 "J index of a test point to check ice relaxation dumped to log file")
+
+  ! Read in relaxation rate, s-1, for ice thickness and partial area
+  filename = trim(inputdir)//trim(relaxrate_file)
+  call log_param(param_file, mdl, "INPUTDIR/ISPONGE_RELAX_FILE", filename)
+  if (.not.file_exists(filename, G%Domain)) &
+    call SIS_error(FATAL, " initialize_icerelax_file: Unable to open "//trim(filename))
+
+  call MOM_read_data(filename, rlxrate_var, Irelax(:,:), G%Domain, scale=US%s_to_T)
+  max_rlxrate =  maxval(Irelax*US%T_to_s)
```
Hallberg-NOAA: Please move lines 176 and 177 setting `max_rlxrate` and the time variables into the `if (verb_msg > verbosity)` block, as they are not used otherwise.

src/SIS_sponge.F90
```diff
@@ -0,0 +1,991 @@
...
+
+  CS%time_varying_sponges = .true.  ! TODO: add option to SIS_input for not time varying rlx fields
+  CS%spongeDataOngrid = .true.
+  if (present(time_var_rlx)) CS%time_varying_sponges = time_var_rlx
+  if (present(sponge_ongrid)) CS%spongeDataOngrid = sponge_ongrid
+
+  CS%use_isponge = use_isponge
+  if (present(itest) .and. present(jtest)) then
+    write(mesg,'(A," itest/jtest =",2(i5,1x))') trim(mdl), itest, jtest
+    write(*,'(A)') trim(mesg)
```
Hallberg-NOAA: Using `write(*,...)` from all PEs can generate a very large volume of redundant information.  Instead, use `SIS_mesg(msg)` here to only write this message from the root PE.

src/SIS_sponge.F90
```diff
@@ -0,0 +1,991 @@
...
+
+  ! local variables
+  integer :: isd, ied, jsd, jed      !< data domain indices
+  integer :: isdG, jsdG              !< Global indx, start pnts. data domain
+  integer :: m, i, j, k, col
+  integer :: iiG, jjG
+  integer :: CatIce                  !< number of sea ice categories.
+  integer :: icat0                   !< cat where to assign original hice and cice to begin distr.
+
+  real, allocatable, dimension(:,:) :: cice2d, hice2d
```
Hallberg-NOAA: Please add comments describing each of the real variables in this routine, along with their units.

src/SIS_sponge.F90
```diff
@@ -0,0 +1,991 @@
...
+  if (allocated(ccat)) deallocate(ccat)
+  if (allocated(hcat)) deallocate(hcat)
+  if (allocated(volcat)) deallocate(volcat)
+
+end subroutine distribute_ice2cats
+
+!< Check if total ice thkn*conc and conc are conserved (i.e. equal original hice, cice)
+subroutine check_hcice(CatIce, hcat, ccat, hice, cice, err_hice, err_cice, &
+                       hice_tot, cice_tot, iG, jG, str, verb)
+  integer, intent(in) :: CatIce
```
Hallberg-NOAA: SIS2 mostly follows the MOM6 style guide, so all subroutine arguments are required to be described in doxygen syntax comments, including units for real variables.  This same comment applies to several of the following routines.

src/SIS_sponge.F90
```diff
@@ -0,0 +1,991 @@
...
+use SIS2_ice_thm,      only : enth_from_TS, Temp_from_En_S, enthalpy_liquid, calculate_T_freeze
+use SIS_optics,        only : VIS_DIR, VIS_DIF, NIR_DIR, NIR_DIF    ! debugging only, delete later
+use SIS_utils,         only : is_NaN
+
+implicit none; private
+
+#include <SIS2_memory.h>
+
+public initialize_icerelax_file, apply_isponge, set_up_isponge_field, SIS_sponge_end
+public global_to_local_ij
```
Hallberg-NOAA: I do not believe that `global_to_local_ij` is used outside of this module. If so it need not be declared as public.

src/SIS_sponge.F90
marshallward: @Hallberg-NOAA I believe this reports on an individual grid point so would only be printed from a single PE.

(But it probably can't hurt to use `SIS_mesg`.)

src/SIS_sponge.F90
```diff
@@ -0,0 +1,991 @@
...
+      select case (trim(CS%var(m)%fld_name))
+        case('mH_ice')
+          hcat(1:CatIce) = CS%Ref_val(m)%p(col,1:CatIce)*Iscale
+        case('part_size')
+          ccat(1:CatIce) = CS%Ref_val(m)%p(col,1:CatIce)
+      end select
+    enddo
+
+    call partial_area_total(CatIce, ccat(1:CatIce), cice_tot)
+    call ice_thkn_total(CatIce, ccat(1:CatIce), hcat, hice_tot)
```
marshallward: `ccat` is allocated as size `CatIce` in L704, so using `ccat(1:CatIce)` seems redundant.  This could also cause an implicit copy inside of the function.  Can `ccat` be used here?

marshallward:
Hi @DmitryDukhovskoy
Several of the functions (`find_icat`, `partial_area_total`, `ice_thnk_total`, `check_hice`) were written with a size argument `CatIce`. This could be removed by passing the arguments as deferred size, e.g.
```
real, intent(in) :: cice_cat(:)
```
and using `size()` to set the loop bounds.

Not urgent, but if you are making other revisions then this could be included.

src/SIS_slow_thermo.F90
DmitryDukhovskoy: Thank you, missed that. Done.

src/SIS_sponge.F90
DmitryDukhovskoy: Corrected

src/SIS_sponge.F90
DmitryDukhovskoy: Corrected.

src/SIS_sponge.F90
DmitryDukhovskoy: Good observation, max_rlxrate and get_date(Time, ...) moved into the block.  Also get_time(Time, start_of_day, num_days) is not needed and has been deleted.

src/SIS_sponge.F90
DmitryDukhovskoy: Comments have been added.

src/SIS_sponge.F90
DmitryDukhovskoy: The description has been added

src/SIS_sponge.F90
DmitryDukhovskoy: This is correct, I also found this while working on other changes in the code. get_time is not needed.

src/SIS_sponge.F90
DmitryDukhovskoy: I totally agree — I also cringed when I used write(*,...) for debugging. However, I couldn’t come up with a better solution at the time. Marshall is right: the message is reported only for a specific grid point on a single PE. I did try using SIS_mesg, but it didn’t work. I’ll give it another shot. Happy to implement a better approach if you have any suggestions.

src/SIS_sponge.F90
DmitryDukhovskoy: Corrected

src/SIS_sponge.F90
DmitryDukhovskoy: Marshall, you are probably right. I noticed this is the way it is done in several SIS2 codes, e.g. 
https://github.com/DmitryDukhovskoy/SIS2_irelax/blob/2c49005d1d7132776fff9b501cabe6ae85ef5f57/src/ice_ridge.F90#L309
The issue is that some arrays are allocated as (0:CatIce) while others use (1:CatIce). It's very easy to make a mistake and assign the wrong arrays. I encountered this myself when I used ccat = IST%part_size(i,j,:), where part_size was declared as (0:CatIce). It took me a while to figure out why the relaxation wasn't working—there was no compilation error.

Although it's a bit redundant, explicitly specifying dimensions makes the code safer. That’s why several SIS codes follow this approach. I'm happy to change the code back (as I had it before and as you're suggesting) if that is the preferred style.

DmitryDukhovskoy:
> Hi @DmitryDukhovskoy Several of the functions (`find_icat`, `partial_area_total`, `ice_thnk_total`, `check_hice`) were written with a size argument `CatIce`. This could be removed by passing the arguments as deferred size, e.g.
> 
> ```
> real, intent(in) :: cice_cat(:)
> ```
> 
> and using `size()` to set the loop bounds.
> 
> Not urgent, but if you are making other revisions then this could be included.

Hi @marshallward , thank for the suggestion. It is done.

src/SIS_sponge.F90
marshallward: You have a good point, the start index of ice category changes often enough that it's worth calling it out.

I am mainly concerned that it could trigger a redundant copy, e.g.
```
tmp(:) = ccat(1:CatIce)
call partial_area_total(CatIce, tmp, cice_tot)
```
It's hard to know what a compiler will do in this case.  If the input is `intent(in)` and it knows that `CatIce` won't change, then the code output is probably identical.  But I would feel more comfortable if it were not given the option to copy.

Another way to emphasize the one-indexing is to make it explicit in the allocation?
```
allocate(ccat(1:CatIce))
```
This information could also be provided in the variable description.

DmitryDukhovskoy: I got it, you are right. I'll make changes in the code. Thanks!

DmitryDukhovskoy:
Hi @Hallberg-NOAA and @marshallward, I think I've addressed all your comments in the recent commit, except for Bob's comment regarding write(*,'(A)') trim(mesg) for the test point. I could not come up with a better solution. I'm ok to remove the "test point" option from the code to resolve this problem. The "test point" was very helpful during code development but perhaps it is not so useful for the user. Let me know. I compiled the updated code and ran several test simulations to verify the code performance. See attached figure showing ice thickness fields from the simulations with varying relaxation rates and the target fields (at the top).  Thank you for your suggestions and time. 
Screenshot 2025-05-02 at 3 04 40 PM

src/SIS_sponge.F90
```diff
@@ -0,0 +1,973 @@
...
+                 fail_if_missing=.true.)
+  call get_param(param_file, mdl, "ISPONGE_STATE_FILE", state_file, &
+                 "The name of the file with the state to relax toward.", &
+                 fail_if_missing=.true.)
+  call get_param(param_file, mdl, "ISPONGE_ITHCK_VAR", ithck_var, &
+                 "The name of the ice thickness variable in "//&
+                 "ISPONGE_STATE_FILE.", default="ithkn")
+  call get_param(param_file, mdl, "ISPONGE_IAREA_VAR", iarea_var, &
+                 "The name of the ice partial area/concentration variable in "//&
+                 "ISPONGE_STATE_FILE.", default="ithkn")
```
theresa-cordero: Should the default for `ISPONGE_IAREA_VAR` be different than the default for `ISPONGE_ITHCK_VAR`?

DmitryDukhovskoy: It should, good catch. Thank you. I modified the code, made sure it compiled, committed and pushed the changes.

src/SIS_sponge.F90
Hallberg-NOAA: Try using `call SIS_mesg(mesg, all_print=.true.)`  Without the `all_print` argument, only the root PE writes this message.

DmitryDukhovskoy: Hi @Hallberg-NOAA, it seemed like a right approach, unfortunately it didn't work.  I implemented this change in the code and all messages disappeared. Then I also tried doing this:   

  CS%use_isponge = use_isponge
  if (present(itest) .and. present(jtest)) then
    write(mesg,'(A," itest/jtest =",2(i5,1x))') trim(mdl), itest, jtest  
    write(*,'(A)') trim(mesg)             ! <--- Old version
    call SIS_mesg(trim(mesg), all_print=.true.)  ! <--- Suggested change

In the log file, I'm getting the message only from my version and none from SIS_mesg. 
initialize_isponge itest/jtest =   13     6

I also tried this to see if all_print works:
  write(mesg,'(A,": SIS_SPONGE IS ON")') trim(mdl)
  call SIS_mesg(mesg, all_print=.true.)

Checking the MOM_mesg subroutine, I would expect to see this message reported from all PEs, however, I'm getting only one message:
NOTE from PE     0: initialize_isponge: SIS_SPONGE IS ON

I'm not sure why all_print option does not work, it should.

src/SIS_sponge.F90
marshallward: I looked into this a little more, and I was being overly cautious.  `ccat(1:CatIce)` is considered "simply contiguous", regardless of the size of `ccat`, which ensures that it will not produce a temporary copy.

Sorry about making this an issue.  I guess we're all still learning every day :flushed: But thank you for making the change.

marshallward:
@DmitryDukhovskoy We've had a fair bit of debate on our end about the `[ij]test` feature, and how to integrate into the codebase.  It looks like it has been an invaluable tool during development of the SIS sponge, and a lot of careful work has been put into its design.  But there are problems we need to consider.

* `write()` is a problematic command for a few reasons:

  * It lacks parallel control.  There is the burden of working out which rank is the I/O source.

  * Generally we cede I/O to FMS, rather than doing directly (although exceptions can be found).  It is possible to get a `stdout` unit from FMS, but this is typically only done at the driver level.

  * `write()` on GPU would raise an error, if we ever get to that point.

* Several `if (CS%itest > 0)` tests are within do-loops, which can impair vectorization and the overall performance of model loops.

While I imagine that we could work through these problems, it could take time to reach a point which satisfies everyone.

I know it's probably not what you would like to hear, but how receptive would you be to removing the `[ij]test` feature?

src/SIS_sponge.F90
DmitryDukhovskoy: Hi @marshallward, no worries. I suppose we want to be conservative with code development. I wasn’t thinking about temporary copies when I originally wrote this, and your suggestion seems reasonable.

DmitryDukhovskoy:
> @DmitryDukhovskoy We've had a fair bit of debate on our end about the `[ij]test` feature, and how to integrate into the codebase. It looks like it has been an invaluable tool during development of the SIS sponge, and a lot of careful work has been put into its design. But there are problems we need to consider.
> 
> * `write()` is a problematic command for a few reasons:
>   
>   * It lacks parallel control.  There is the burden of working out which rank is the I/O source.
>   * Generally we cede I/O to FMS, rather than doing directly (although exceptions can be found).  It is possible to get a `stdout` unit from FMS, but this is typically only done at the driver level.
>   * `write()` on GPU would raise an error, if we ever get to that point.
> * Several `if (CS%itest > 0)` tests are within do-loops, which can impair vectorization and the overall performance of model loops.
> 
> While I imagine that we could work through these problems, it could take time to reach a point which satisfies everyone.
> 
> I know it's probably not what you would like to hear, but how receptive would you be to removing the `[ij]test` feature?

@marshallward, thanks for taking the time to work on this problem. These are all important concerns. I like this feature, but it’s probably not that critical for the user unless ice nudging needs debugging. I'm ok to remove the ij-test feature. I'll make the necessary changes in the code.

DmitryDukhovskoy:
Hi, The updated version of the code with the ij-test debugging feature removed has been pushed. I’ve tested this version, and everything appears to be working correctly; I haven’t encountered any issues so far. I do regret having to remove the ij-test feature, as I relied on it heavily while evaluating recent code performance and resolving issues. However, I understand the complications it introduced and why it needed to be removed from the shared version.
For now, I’ll maintain a separate version with the ij-test functionality for my own use. 
Thanks!

marshallward (dismissed):
@DmitryDukhovskoy Thanks for your understanding.

I can see the value in a per-element diagnostic, it feels like a gap in our existing tools.  But it may be easier to evaluate as a separate PR.

I think all of the issues have been addressed so I will run this through the regression suite and, assuming it passes, will merge it.

marshallward:
Gaea regression: https://gitlab.gfdl.noaa.gov/ogrp/mom6ci/SIS2/-/pipelines/27659 :heavy_check_mark: :yellow_circle:

marshallward (approved):

PR #219, opened 2025-06-18
Add option to neglect massless snow and or pond during continuity updates and compress_ice

MJHarrison-GFDL:
This adds a new parameter **NEGLECT_MASSLESS_SNOW_POND** which is defaulted to False in order to retain existing answers, but should be set to True in order to avoid roundoff level snow and or pond mass contributing to subsequent errors in SIS_continuity.


theresa-cordero:
Matt, can you modify this branch to be just the changes from commit `4171045 `?
Right now we just have the hard coded option to always neglect snow and ponds but that seems to be causing some problems in certain configurations where previously negative snow and ponds were taken into account. 

I'd like to merge in your fix which includes a flag for this.

PR #220, opened 2025-06-18
+Rename checksum scale argument to unscale

Hallberg-NOAA:
Renamed the `scale` argument to the various checksum calls used by SIS2 to `unscale`, and added `unscale` optional arguments to the various `check_redundant()` routines.  Both of these changes bring SIS2 into line with what is used now in MOM6 to facilitate the tighter integration of SIS2 and MOM6.  All calls to checksums had the name of the scale optional argument changed, and several calls to the `check_redundant()` routines now have `unscale` arguments.  In addition, descriptions of the units of about 55 real SIS2 variables were added, mostly by copying comments directly from the corresponding MOM6 routines.  All answers are bitwise identical, but there is a change in the name of an optional argument to a number of publicly visible routines.

### DIFF
--- a/src/SIS_debugging.F90
+++ b/src/SIS_debugging.F90
@@ -30,6 +30,7 @@ implicit none ; private
 
 public :: check_redundant_C, check_redundant_B, check_redundant_T
 public :: SIS_debugging_init
+public :: query_debugging_checks
 
 ! These interfaces come from MOM_checksums.
 public :: hchksum, Bchksum, is_NaN, chksum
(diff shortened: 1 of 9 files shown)


theresa-cordero (approved):

MitchBushuk (approved):
I agree with Theresa's assessment, and this is ready to be approved.

marshallward:
Gaea regression: https://gitlab.gfdl.noaa.gov/ogrp/mom6ci/SIS2/-/pipelines/28666 :heavy_check_mark: :yellow_circle: 

Parameters are out of date, but this could be from a previous PR.

PR #221, opened 2025-07-03
Fixed a bug that could incorrectly trigger a warning about non-conserved ice properties in SIS_sponge code

DmitryDukhovskoy:
Fixed a bug that could incorrectly trigger a warning about not conserved aggregated ice thickness or concentration during the distribution routine when input target fields are distributed into ice thickness categories.  Changed verbosity check "if" statement that appeared to be flipped. Minor editing of the comments.

### DIFF
--- a/src/SIS_sponge.F90
+++ b/src/SIS_sponge.F90
@@ -114,7 +114,7 @@ subroutine initialize_icerelax_file(param_file, G, IG, CS, US, IST, Time)
                                               !! mean grid cell value  kg/m [R Z L ~> kg m-1]
   real, allocatable, dimension(:,:) :: rlx_C  !< A temporary array for reading relax target ice partial area
 
-  integer, parameter :: verb_msg = 9 !< verbosity level for messages
+  integer, parameter :: verb_msg = 8 !< verbosity level for messages
   integer :: i, j, k, is, ie, js, je, ncat
   integer :: isd, ied, jsd, jed
   integer :: isc, iec, jsc, jec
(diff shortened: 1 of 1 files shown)


theresa-cordero (approved):
These changes are reasonable to me. Since this code is not used by a large number of people and in a published simulation I do not think there is a need to preserve the old answers.

src/SIS_sponge.F90
```diff
@@ -597,23 +606,22 @@ subroutine distribute_ice2cats(CS, IG, G, scaled, eps_err)
...
+    hice = hice2d(i,j)    !< in input units [m3*m-2], also equivallent to cell-mean ice thickness
+    cice = cice2d(i,j)    !< total partial area (conc)
+    if (hice < 1.e-10 .or. cice < 1.e-10) then
       hice = 0.0 ; cice = 0.0
       do m=1,CS%fldno ; do k=1,CatIce
         select case (trim(CS%var(m)%fld_name))
           case('mH_ice')
-            CS%Ref_val(m)%p(col,k) = IG%mH_cat_bound(k+1) !< The lower mass/unit area limits 
-                                                          !! for ice cat [R Z ~> kg m-2].
+            CS%Ref_val(m)%p(col,k) = 0.0
```
theresa-cordero, line 624: Dmitry, why did you change this from `IG%mH_cat_bound(k+1)` to `0.0`?

DmitryDukhovskoy: @theresa-cordero Theresa, this change is inside an if block that checks whether the target ice concentration or thickness is near zero, indicating an open water grid cell. It should not affect the model output and was made primarily as a safeguard to prevent potential errors when the input thickness is smaller than the value assigned to IG%mH_cat_bound(k+1). The thickness is set to 0 and open water to 1, the do block goes to next grid cell.

theresa-cordero:
Dmitry and I agree that previous answers with sponges do not need to be saved. Answers should only change for cases with ice sponges turned on.

Hallberg-NOAA:
This PR has passed SIS2 pipeline testing at https://gitlab.gfdl.noaa.gov/ogrp/mom6ci/SIS2/-/pipelines/28878.

PR #226, opened 2025-10-25
 Adding pond ice, prep for new sealvl ponds

kshedstrom:
- Not actually calling Icepack ponds yet.

### DIFF
--- a/src/SIS_transport.F90
+++ b/src/SIS_transport.F90
@@ -84,9 +84,11 @@ end type SIS_transport_CS
 type, public :: cell_average_state_type ; private
   real, allocatable, dimension(:,:,:) :: m_ice  !< The mass of ice in each thickness category
                                                 !! per unit total area in a cell [R Z ~> kg m-2].
-  real, allocatable, dimension(:,:,:) :: m_snow !< The mass of ice in each thickness category
+  real, allocatable, dimension(:,:,:) :: m_snow !< The mass of snow in each thickness category
                                                 !! per unit total area in a cell [R Z ~> kg m-2].
   real, allocatable, dimension(:,:,:) :: m_pond !< The mass of melt pond water in each thickness
+                                                !! category per unit total area in a cell [R Z ~> kg m-2].
+  real, allocatable, dimension(:,:,:) :: m_pond_ice !< The mass of melt pond ice in each thickness
                                                 !! category per unit total area in a cell [R Z ~> kg m-2].
   real, allocatable, dimension(:,:,:) :: mH_ice !< The mass of ice in each thickness category
                                                 !! per unit of ice area in a cell [R Z ~> kg m-2].  The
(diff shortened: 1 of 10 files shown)


src/ice_pond.F90
```diff
@@ -0,0 +1,513 @@
...
+
+end subroutine ice_ponds_init
+
+!> ice_ponds is a wrapper for the Icepack pond routines
+subroutine ice_ponds(IST, G, IG, mca_ice, mca_snow, mca_pond, TrReg, CS, US, dt, &
+                       rdg_rate, rdg_height)
+  type(ice_state_type),              intent(inout) :: IST !< A type describing the state of the sea ice.
+  type(SIS_hor_grid_type),           intent(inout) :: G   !< G The ocean's grid structure.
+  type(ice_grid_type),               intent(inout) :: IG  !< The sea-ice-specific grid structure.
+  real, dimension(SZI_(G),SZJ_(G),SZCAT_(IG)), intent(inout) :: mca_ice  !< mass of ice?
```
Hallberg-NOAA, line 108: Please specify the units of all real arguments and internal variables using the same syntax (here it would probably be [R Z ~> kg m-2]) as is used elsewhere in the SIS2 code.  This comment applies to many places in ice_pond.F90.

Hallberg-NOAA (dismissed):
I have looked through these proposed changes, and they look like a valuable step towards adding pond ice to SIS2.  The only thing I would change first is to add some more documentation of the units of real variables (and perhaps some other syntax changes) to align with the SIS2/MOM6 style guide at https://github.com/NOAA-GFDL/MOM6/wiki/Code-style-guide.

src/ice_pond.F90
Hallberg-NOAA: Although I am reluctant to take in code that does not follow the style guide, in this case I think that we can safely accept this code as-is and then fix the documentation of the units of the newly introduced real variables later.

Hallberg-NOAA (dismissed):
I have decided that in this particular circumstance, we can accept the code as-is as a valuable contribution, despite missing documentation, and add the documentation of the variable units later.

src/SIS_transport.F90
```diff
@@ -1183,7 +1238,8 @@ subroutine cell_mass_from_CAS(CAS, G, IG, mca, scale)
 
   do j=jsc,jec ; do i=isc,iec ; mca(i,j) = 0.0 ; enddo ; enddo
   do k=1,nCat ; do j=jsc,jec ; do i=isc,iec
-    mca(i,j) = mca(i,j) + H_to_units * (CAS%m_ice(i,j,k) + (CAS%m_snow(i,j,k) + CAS%m_pond(i,j,k)))
+    mca(i,j) = mca(i,j) + H_to_units * ((CAS%m_ice(i,j,k) + CAS%m_snow(i,j,k)) + \
```
marshallward: `\` should probably be `&`.  This is failing on the NVIDIA compilers (and none of the others!!).

Hallberg-NOAA: We think that the final `\` is being treated by CPP as a line continuation token and the next line is being merged with this one for the intel and gnu compilers.  We do not know why Nvidia is not doing this, but still this is non-standard Fortran so we should avoid using this construct.

marshallward: Looks like this is defined CPP behavior, and NVIDIA is not in compliance.  Still happening in 26.1
```
$ cat x.F90 
x = 1 + \
2
end
```
error
```
$ nvfortran x.F90 
NVFORTRAN-W-0025-Illegal character (\) - ignored (x.F90: 1)
NVFORTRAN-S-0034-Syntax error at or near end of line (x.F90: 1)
NVFORTRAN-S-0018-Illegal label 2 - must be followed by a keyword or an identifier (x.F90: 2)
  0 inform,   1 warnings,   2 severes, 0 fatal for MAIN
```
I hesitate to call this a bug, since I don't think compilers are obliged to exactly do what is in the CPP docs, but nonetheless we should probably avoid it 😅

Hallberg-NOAA:
This PR has passed pipeline testing at https://gitlab.gfdl.noaa.gov/ogrp/mom6ci/SIS2/-/pipelines/30336.

Hallberg-NOAA (approved):
This PR is now passing all of our tests and is ready to be merged in.  Thank you for this contribution, and for your patience as we dealt with this PR, @kshedstrom!

PR #229, opened 2025-12-17
+Enhance SIS_diag_mediator to mimic MOM6

Hallberg-NOAA:
This commit greatly enhances the capabilities of the SIS2 diag_mediator by pulling in multiple capabilities that had been added to the MOM6 diag_mediator in the 13 years since the SIS2 diag_mediator was spawned from it.  The new capabilities include the ability to handle CMOR name options for diagnostics and related metadata, write out scalar diagnostics, record cell methods as metadata, use integer index axes in place of nominal latitudes and longitudes, print out checksums of all primary diagnostics and detect any obsolete diagnostics in diag tables.

  The new public interfaces with this commit include `register_scalar_field()`, `define_axes_group()`, `diag_masks_set()`, `diag_register_area_ids()` and `found_in_diagtable()`.  None of these are exercised as yet by the SIS2 code, but new options that exercise them will appear in future commits.  Also the publicly visible but opaque `axesType` has been renamed to `axes_grp`, mirroring MOM6.  This commit adds the new runtime parameters `USE_INDEX_DIAGNOSTIC_AXES`, `DIAG_MISVAL`, `DIAG_AS_CHKSUM` and `CHKSUM_DIAG_FILE` to support the new capabilities.

  There is a new non-optional `unit_scale_type` argument to `SIS_diag_mediator_init()` and the optional `mask` arguments to `post_data_2d()` and `post_data_3d()` that were logical arrays have been replaced by real arrays, following the pattern used in MOM6.  This has required minor code changes in 4 places outside of SIS_diag_mediator.F90.

  For now, all diagnostic and available_diags files written by SIS2 are identical to what the previous version wrote out, but there is commented out code that will add new metadata and formatting like that generated by the MOM6 diag_mediator.  It is the nature of this commit that the changes swept up by it are extensive, but with such extensive refactoring this is probably unavoidable and it seems like there would be minimal purpose served by breaking this up into intermediate waypoints.

### DIFF
--- a/src/SIS_diag_mediator.F90
+++ b/src/SIS_diag_mediator.F90
@@ -4,56 +4,105 @@ module SIS_diag_mediator
 ! This file is a part of SIS2. See LICENSE.md for the license.
 
 use ice_grid,               only : ice_grid_type
+use MOM_checksums,          only : chksum0, hchksum, uchksum, vchksum, Bchksum
 use MOM_coms,               only : PE_here
-use MOM_diag_manager_infra, only : diag_manager_init=>MOM_diag_manager_init
+use MOM_cpu_clock,          only : cpu_clock_id, cpu_clock_begin, cpu_clock_end
+use MOM_cpu_clock,          only : CLOCK_MODULE, CLOCK_ROUTINE
+use MOM_diag_manager_infra, only : diag_manager_init=>MOM_diag_manager_init, MOM_diag_manager_end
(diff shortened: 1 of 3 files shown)


adcroft (commented):
A lot of code changed here but it all looks familiar having been mostly copied from MOM6.

Testing at https://gitlab.gfdl.noaa.gov/ogrp/mom6ci/SIS2/-/pipelines/29636

adcroft (approved):
Passed. We need to update the doc files after this.

Issue #230, opened 2025-12-22
Why does SIS2 report a negative zero maximum CFL?


Hallberg-NOAA:
For some cases, SIS2 is reporting a maximum CFL of -0.000 at the first timestep (when all velocities are zero) for many of our test cases with the gnu and Nvidia compilers.  This is easy enough to trap and fix in SIS_sum_output.F90 (with a line like `if (max_CFL == 0.0) max_CFL = abs(maxCFL)`, but I am very curious to know why these negative zeros are coming out of the FMS `mpp_max()` routine (which we wrap with `max_across_PEs()` in `MOM6/config_src/infra/FMS[12]/MOM_coms_infra.F90`.  Curiously, this does not appear to impact MOM6.

PR #231, opened 2025-12-22
*Fix negative zero maximum CFL

Hallberg-NOAA:
SIS2 reports the global maximum CFL in the seaice.stats files, but in the first timestep when all velocities are zero, it is reporting a maximum CFL of -0.000 with the gnu and Nvidia compilers.  This single-line commit replaces any negative zero maximum CFL returned by `max_across_PEs()` with its absolute value so that only non-negative maximum CFL values are returned.  All solutions are bitwise identical, but this does change the entries in the seaice.stats files so it may appear to fail automated regression testing in some cases.

  This commit addresses the problem in SIS2 that is noted in https://github.com/NOAA-GFDL/SIS2/issues/230, but it does not address the underlying question of why this bizarre behavior is happening in the first place.

### DIFF
--- a/src/SIS_sum_output.F90
+++ b/src/SIS_sum_output.F90
@@ -438,6 +438,7 @@ subroutine write_ice_statistics(IST, day, n, G, US, IG, CS, message, check_colum
     enddo ; enddo
   endif
   call max_across_PEs(max_CFL)
+  if (max_CFL == 0.0) max_CFL = abs(max_CFL) ! This is to deal with negative zeros.
 
   ! Set combinations of scalign factors that rescale back to MKS units for output
   area_scale = US%L_to_m**2


marshallward (approved):
As far as I know, the MPI library can strip the sign of zero during any reduction, making it impossible to reproduce signed zeros, so explicit assignment to (positive) zero is the only way to reliably reproduce the sign.

Normally I would say that this is better handled at the `MOM_coms` level.  But the delayed adoption of MOM6 code in SIS2 justifies taking this right away.

The only minor improvement here would be to replace `max_CFL = abs(max_CFL)` with `max_CFL = 0.`.  But I will approve as it is, and this line can be removed after a future update to `max_across_PEs`.

Hallberg-NOAA:
This PR completed the pipeline testing at https://gitlab.gfdl.noaa.gov/ogrp/mom6ci/SIS2/-/pipelines/29802 with the expected failure due to the changes in some seaice.stats files that were previously reporting negative zero maximum CFL numbers.

PR #233, opened 2025-12-22
Fix syntax in 84 unit descriptions

Hallberg-NOAA:
Corrected the syntax of 83 unit descriptions used in diagnostic registrations in 7 files and explicitly name the units arguments in 101 calls to `register_SIS_diag_field()`.  Also added a missing unit argument to one call to `register_restart_field()`.  All answers are bitwise identical, but there are metadata changes in some diagnostic and restart files.

PR #234, opened 2025-12-22
Set range and units for 19 nondimensional diagnostics

Hallberg-NOAA:
Set range and corrected the units for 19 nondimensional diagnostics and explicitly named the units arguments in these calls.  Also changed the descriptions of the range arguments to routines in `SIS_diag_mediator.F90` to no longer state that these arguments are unused in SIS2.  All answers are bitwise identical, but there are metadata changes in some diagnostic files,


MitchBushuk (approved):
These changes are straightforward and all look correct. I have checked the three files that were changed.

Hallberg-NOAA:
This PR has passed pipeline testing at https://gitlab.gfdl.noaa.gov/ogrp/mom6ci/SIS2/-/pipelines/29826 with the expected warnings about changes to diagnostic units (plus warnings about some similar changes from PR #233).

PR #235, opened 2025-12-22
*Correct metadata or calculation of 19 diagnostics

Hallberg-NOAA:
Corrected the longname, unit description, conversion factor or the calculation of 19 diagnostics in 4 files.  In two cases, the calculations were revised to avoid array syntax calculation of whole arrays, including potentially uninitialized halo regions.  In the case of two other diagnostics, necessary unit conversion factors were missing altogether.  In yet other cases, the unit conversion factors were moved into the `register_SIS_diagnostics()` conversion factors to facilitate comparison with the declared units.  All solutions are bitwise identical, but the metadata associated with some diagnostics have been changed, and some diagnostics that previously were not passing unit testing are passing now.

### DIFF
--- a/src/SIS_ice_diags.F90
+++ b/src/SIS_ice_diags.F90
@@ -86,6 +86,10 @@ subroutine post_ice_state_diagnostics(IDs, IST, OSS, IOF, dt_slow, Time, G, US, 
     temp_snow   ! A diagnostic array with the snow temperature [C ~> degC].
   real, dimension(SZI_(G),SZJ_(G),IG%CatIce) :: &
     rdg_frac    ! fraction of ridged ice per category [nondim]
+  real, dimension(SZI_(G),SZJ_(G),IG%CatIce) :: &
+    mass_by_cat ! Sea ice mass per unit ocean area by thickness category [R Z ~> kg m-2]
+  real, dimension(SZI_(G),SZJ_(G),IG%CatIce) :: &
+    thick_by_cat ! Sea ice thickness by thickness category [Z ~> m]
   real, dimension(SZI_(G),SZJ_(G)) :: diagVar ! A temporary array for diagnostics.
   real, dimension(IG%NkIce) :: S_col ! Specified thermodynamic salinity of each
                                      ! ice layer if spec_thermo_sal is true [S ~> gSalt kg-1]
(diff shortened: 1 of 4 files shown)


src/SIS_transport.F90
```diff
@@ -1268,8 +1267,7 @@ subroutine SIS_transport_init(Time, G, IG, US, param_file, diag, CS, continuity_
 !  CS%id_rdgo = register_diag_field('ice_model', 'RDG_OPEN', diag%axesT1, Time, &
 !               'rate of opening due to ridging', '1/s', conversion=US%s_to_T, missing_value=missing)
 !  CS%id_rdgv = register_diag_field('ice_model', 'RDG_VOSH', diag%axesT1, Time, &
-!               'volume shifted from level to ridged ice', 'm^3/s', conversion=US%RZ_T_to_kg_m2s*US%L_to_m**2, &
-!                missing_value=missing)
+!               'Mass shifted from level to ridged ice', units='kg s-1', conversion=US%RZ_T_to_kg_m2s*US%L_to_m**2)
```
MitchBushuk, line 1270: This diagnostic does not exist yet, but I agree that tracking ice mass changes, as opposed to volume, would be more useful. Given this, it seems that we should also change the diagnostic name `RDG_VOSH` which currently has a `VO` for volume. Potentially the new name could be `RDG_MISH`, where the `MI` stands for Mass of Ice.

Hallberg-NOAA: The name of this candidate diagnostic has now been changed from `RDG_VOSH` to `RDG_MISH` as suggested.

src/SIS_ice_diags.F90
```diff
@@ -180,15 +196,15 @@ subroutine post_ice_state_diagnostics(IDs, IST, OSS, IOF, dt_slow, Time, G, US,
     call post_data(IDs%id_ext, diagVar, diag)
   endif
   if (IDs%id_hp>0) call post_avg(IDs%id_hp, IST%mH_pond, IST%part_size(:,:,1:), & ! mw/new
-                                 diag, G=G, scale=US%RZ_to_kg_m2/1e3, wtd=.true.) ! rho_water=1e3
+                                 diag, G=G, scale=1.0/(1e3*US%kg_m3_to_R), wtd=.true.) ! rho_water=1e3 [kg m-3]
   if (IDs%id_hs>0) call post_avg(IDs%id_hs, IST%mH_snow, IST%part_size(:,:,1:), &
-                                 diag, G=G, scale=US%Z_to_m/Rho_snow, wtd=.true.)
+                                 diag, G=G, scale=1.0/Rho_snow, wtd=.true.)
```
MitchBushuk, line 201: Should Rho_snow here and Rho_ice below also have rescaling factors applied, analogous to id_hp above (Line 199)?

Hallberg-NOAA: No, these extra rescaling factors are are not necessary because Rho_snow and Rho_ice are already in rescaled units of [R ~> kg m-3], with the rescaling occurring where the runtime parameters RHO_SNOW and RHO_ICE are read in within `ice_thermo_init()` at about lines 1542 and 1545 of `SIS2_ice_thm.F90`.  Line 199 needed this rescaling factor because we are using a hard-coded density of pond water in mks units of [kg m-3].

MitchBushuk: Thank you for the clarification on this point.

MitchBushuk (approved):

Hallberg-NOAA:
This PR has passed pipeline testing at https://gitlab.gfdl.noaa.gov/ogrp/mom6ci/SIS2/-/pipelines/29858 with the expected warnings about changes in the documented units of diagnostics.

PR #236, opened 2025-12-25
+Revise SIS2 available diags file to be like MOM6

Hallberg-NOAA:
Revised the SIS2 diag mediator code that creates the SIS.available_diags file to follow the same pattern of logging as MOM6.  The specific changes include moving the module name(s) to their own line, adding static diagnostics to the variables that are logged in the available_diags file, logging the axes of the diagnostic, and adding cell metrics to describe how a diagnostic might be reduced and logging these as well.  All solutions are bitwise identical, but there are new entries and added information in the SIS.available_diags and SIS_fast.available_diags files.


MitchBushuk (approved):
This is a useful upgrade to the SIS2 available diags file which will provide users with more information than was previously available. It will also make the SIS2 output consistent with the MOM6 format.

Hallberg-NOAA:
This PR has passed pipeline testing at https://gitlab.gfdl.noaa.gov/ogrp/mom6ci/SIS2/-/pipelines/29862 with the expected warnings about changes to the formatting and content of the SIS_available_diags files.

PR #237, opened 2025-12-25
+Change some SIS2 axis units to CMIP standards

Hallberg-NOAA:
Modified `set_SIS_axes_info()` to follow the pattern from MOM6 and take the units for the horizontal axes from the grid type.  In most cases, this has the effect of changing the units from "degrees_E" and "degrees_N" to "degrees_east" and "degrees_north" to follow CMIP protocols, and to mirror changes that were made to MOM6 in 2017.  All solutions are bitwise identical, but there are changes to the metadata in some output files, and some runtime parameters that had been read from both `set_SIS_axes_info()` and `set_grid_metrics()` in MOM6/src/initialization/MOM_grid_initialize.F90 are now only read in the latter routine.  As a result of these changes, `set_SIS_axes_info()` no longer takes a `param_file_type` argument and the grid arguments are now `intent(in)`.

### DIFF
--- a/src/SIS_diag_mediator.F90
+++ b/src/SIS_diag_mediator.F90
@@ -194,10 +194,9 @@ integer :: id_clock_diag_mediator
 contains
 
 !> Set up the grid and axis information for use by SIS.
-subroutine set_SIS_axes_info(G, IG, param_file, diag_cs, set_vertical, axes_set_name)
-  type(SIS_hor_grid_type), intent(inout) :: G   !< The horizontal grid type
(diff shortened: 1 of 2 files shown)


MitchBushuk (dismissed):
This appears to be a useful and benign change. It makes SIS2 consistent with CMIP and MOM6 standards, which is a useful improvement.

Hallberg-NOAA:
This PR has passed pipeline testing at https://gitlab.gfdl.noaa.gov/ogrp/mom6ci/SIS2/-/pipelines/29868.

adcroft (approved):
Beautiful!

PR #239, opened 2025-12-28
+Fix dimensional rescaling of DT_ICE_DYNAMICS

Hallberg-NOAA:
Added the missing dimensional rescaling factor for `DT_ICE_DYNAMICS` in `SIS_tracer_advect_init()`.  This commit could change answers in some case when dimensional rescaling of time is being tested in SIS2, but because this particular timestep is only used to determine the maximum number of iterations for the advection of ice thicknesses and tracers, and the maximum number of iterations is always at least 1, this appears to have gone undetected heretofore in the short tests with small CFL numbers that we have used for dimensional consistency testing in SIS2.  This change includes adding a new `unit_scale_type` argument to this same routine.  All answers are bitwise identical when `T_RESCALE_POWER` is 0 in SIS_input (as is always the case in production runs), but there is a new argument to a publicly visible interface.

### DIFF
--- a/src/SIS_tracer_advect.F90
+++ b/src/SIS_tracer_advect.F90
@@ -1849,10 +1849,11 @@ end subroutine advect_tracers_thicker
 
 !> Initialze allocate the control structure for the SIS_tracer_advect module
 !! and set its parameters
-subroutine SIS_tracer_advect_init(Time, G, param_file, diag, CS, scheme)
+subroutine SIS_tracer_advect_init(Time, G, US, param_file, diag, CS, scheme)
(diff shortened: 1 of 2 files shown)


MitchBushuk (approved):

MitchBushuk:
This is ready to be merged once pipeline testing is complete.

Hallberg-NOAA:
This PR has passed pipeline testing at https://gitlab.gfdl.noaa.gov/ogrp/mom6ci/SIS2/-/pipelines/29951, but with warnings about new diagnostics inherited from the previous PR that is not yet reflected in the latest version of MOM6-examples.

PR #240, opened 2025-12-28
Fix unit scaling in SIS_sponge

Hallberg-NOAA:
Corrected the unit scaling factor for reading the sponge relaxation rate to `US%T_to_s` as is appropriate for a rate.  An incorrect but consistent compensating factor was removed from the expression for `max_rlxrate`, and the line unscaling `dt_slow` into a new variable was also removed.

  There are some other minor changes included in this commit.  The unused variables `I_Nk`, `coeff`, `Idt_slow`, `ice_salin`, `iconc_old`, `ithk_old`, `mask_in` and 6 others were removed from `apply_isponge()` and a three-line expression setting the limited fractional open water category was consolidated down to a single line. The missing rescaling factor for the initial value of `tfi` was added and the initial value of `sice` was change to 0; neither of these initial values appear to be used.  In addition comments describing the units of 23 real variables were added in SIS_sponge.F90.

  All answers are bitwise identical when no dimensional rescaling is applied, but more of the code is now working with rescaled variables.

### DIFF
--- a/src/SIS_sponge.F90
+++ b/src/SIS_sponge.F90
@@ -48,6 +48,7 @@ type, public :: p3d; private
   real, dimension(:,:,:), pointer :: p => NULL() !< A pointer to a 3D array [various]
   character(len=15)               :: fld_name    !< Name of the ice field being relaxed
 end type p3d
+
 !> A structure for creating arrays of pointers to 2D arrays
 type, public :: p2d; private
   type(external_field) :: field !< Time interpolator field handle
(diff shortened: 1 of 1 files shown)


MitchBushuk (approved):
These changes correct the unit rescaling and also make a number of other minor improvements. The changes all look correct to me.

Hallberg-NOAA:
This PR has passed pipeline testing at https://gitlab.gfdl.noaa.gov/ogrp/mom6ci/SIS2/-/pipelines/30853.

PR #228, opened 2025-12-10
Sea ice mass budget diagnostics for CMIP7

William-gregory:
Additional diagnostics for computing sea ice area, sea ice mass, and snow mass budgets. Below are the CMIP diagnostic names how they have been derived.

**Ice concentration**
sidconcdyn : equivalent to XPRTc
sidconcth : equivalent to LSRCc + LSNKc

**Ice mass**
sidmassdyn : equivalent to XPRTi
sidmassevapsubl: split EVAP to derive ice contribution
sidmassgrowthwat : equivalent to FRAZIL/LatHtFus
sidmassgrowthsi: split S2IC to derive ice contribution and average over all part_size categories (incl. open water)
sidmassgrowthbot: equivalent to LSRCi - sidmassgrowthsi - sidmassgrowthwat
sidmassmeltbot : equivalent to BSNKi
sidmassmelttop  : equivalent to LSNKi - BSNKi - sidmassevapsubl
sidmassth : equivalent to LSRCi + LSNKi

**Snow mass**
sisndmassdyn : equivalent to XPRTs
sisndmasssubl: split EVAP to derive snow contribution
sisndmasssi: split S2IC to derive snow contribution and average over all part_size categories (incl. open water)
sisndmassmelt : equivalent to LSNKs - sisndmasssubl - sisndmasssi

Note that, for backwards compatibility, I have kept the output units of the standard mass budget diagnostics (XPRT, LSRC, LSNK, BSNK) as `kgm-2yr-1`

PR #246, opened 2026-01-22
Sea ice mass budget diagnostics for CMIP7 (rebased to eliminate conflicts)

Hallberg-NOAA:
This is a version of PR #228 that has been rebased onto dev/gfdl to address the numerous conflicts that arose because of intervening code changes to make this PR work as intended.  The description from William Gregory of the original PR is as follows: 

  Additional diagnostics for computing sea ice area, sea ice mass, and snow mass budgets. Below are the CMIP diagnostic names how they have been derived.

**Ice concentration**
sidconcdyn : equivalent to XPRTc
sidconcth : equivalent to LSRCc + LSNKc

**Ice mass**
sidmassdyn : equivalent to XPRTi
sidmassevapsubl: split EVAP to derive ice contribution
sidmassgrowthwat : equivalent to FRAZIL/LatHtFus
sidmassgrowthsi: split S2IC to derive ice contribution and average over all part_size categories (incl. open water)
sidmassgrowthbot: equivalent to LSRCi - sidmassgrowthsi - sidmassgrowthwat
sidmassmeltbot : equivalent to BSNKi
sidmassmelttop : equivalent to LSNKi - BSNKi - sidmassevapsubl
sidmassth : equivalent to LSRCi + LSNKi

**Snow mass**
sisndmassdyn : equivalent to XPRTs
sisndmasssubl: split EVAP to derive snow contribution
sisndmasssi: split S2IC to derive snow contribution and average over all part_size categories (incl. open water)
sisndmassmelt : equivalent to LSNKs - sisndmasssubl - sisndmasssi

Note that, for backwards compatibility, I have kept the output units of the standard mass budget diagnostics (XPRT, LSRC, LSNK, BSNK) as [kg m-2 yr-1].

### DIFF
--- a/src/SIS_slow_thermo.F90
+++ b/src/SIS_slow_thermo.F90
@@ -143,7 +143,7 @@ type slow_thermo_CS ; private
   integer :: id_lsrc=-1, id_lsnk=-1, id_bsnk=-1, id_sn2ic=-1
   integer :: id_lsrc_i=-1, id_lsnk_i=-1, id_bsnk_i=-1
   integer :: id_lsrc_s=-1, id_lsnk_s=-1, id_bsnk_s=-1
-  integer :: id_lsrc_a=-1, id_lsnk_a=-1
+  integer :: id_lsrc_c=-1, id_lsnk_c=-1
   !!@}
 end type slow_thermo_CS
 
--- a/src/SIS_slow_thermo.F90
+++ b/src/SIS_slow_thermo.F90
@@ -620,7 +620,9 @@ subroutine SIS2_thermodynamics(IST, dt_slow, CS, OSS, FIA, IOF, G, US, IG)
     bsnk, &               ! The bottom melting mass sink of ice and snow [R Z T-1 ~> kg m-2 s-1]
     bsnk_i, &             ! The bottom melting mass sink of ice [R Z T-1 ~> kg m-2 s-1]
     bsnk_s, &             ! The bottom melting mass sink of snow [R Z T-1 ~> kg m-2 s-1]
-    tmp2d, &              ! A temporary array for mass balance diagnostics [R Z yr-1 ~> kg m-2 yr-1]
+    simass_evap, &        ! The mass of ice evaporated from the surface [R Z T-1 ~> kg m-2 s-1]
+    sisnmass_evap, &      ! The mass of snow evaporated from the surface [R Z T-1 ~> kg m-2 s-1] 
+    tmp2d, &              ! A temporary array for mass balance diagnostics [R Z yr-1 ~> kg m-2 s-1]
     qflx_lim_ice, &       ! Ice limiting heat flux [Q R Z T-1 ~> W m-2]
     qflx_res_ice, &       ! Ice restoring heat flux [Q R Z T-1 ~> W m-2]
     cool_nudge, &         ! A heat flux out of the sea ice that
@@ -682,6 +684,8 @@ subroutine SIS2_thermodynamics(IST, dt_slow, CS, OSS, FIA, IOF, G, US, IG)
   real :: snow_loss      ! The loss of snow mass from transmutation [R Z ~> kg m-2]
   real :: h2o_ice_to_ocn ! The downward water flux from the ice to the ocean [R Z ~> kg m-2]
   real :: h2o_ocn_to_ice ! The upward water flux from the ocean to the ice [R Z ~> kg m-2]
+  real :: evap_i         ! The loss of ice due to evaporation [R Z ~> kg m-2]
+  real :: evap_s         ! The loss of snow due to evaporation [R Z ~> kg m-2]
   real :: evap_from_ocn  ! The evaporation from the ocean [R Z ~> kg m-2]
   real :: bablt       ! The bottom ablation ice and snow loss [R Z ~> kg m-2]
   real :: bablt_i     ! The bottom ablation ice loss [R Z ~> kg m-2]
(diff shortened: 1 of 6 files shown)


Hallberg-NOAA:
Going forward, I would prefer that we not have diagnostics of rates in [kg m-2 yr-1], but instead always use [kg m-2 s-1].  Seconds are well defined, whereas we use at least 4 different definitions of a year in FMS configurations.  For backward compatibility I think that we could always rescale via the diag_table, and there are such large differences in magnitude between the two that there is little risk of confusion.

MitchBushuk:
Thank you @Hallberg-NOAA for your work rebasing this code and resolving conflicts.

 I have tested this code in the OM5 Baltic test case and confirmed that @Hallberg-NOAA's branch `WG_cmip7_diags_rwh` bitwise reproduces solutions from @William-gregory's branch `cmip7_diags`.
 
Below are some notes on the performance and some issues to address (see TO DO below):

**Sea ice mass budgets:**
-The new diagnostics for sea ice mass budgets (`d/dt simass = XPRTi + LSNKi + LSRCi`) close the sea ice mass budget at the gridpoint scale with relative errors of roughly 1e-3. Note that this is the same relative error of the SIS2 sea ice and snow combined mass budget diagnostics (`d/dt MI = XPRT + LSNK + LSRC`).
-`sidmassgrowthwat` is equal to `FRAZIL/LatHtFus` (to float precision)
-Note that the `SN2IC` diagnostic is defined over the ice covered portion of the grid cell. In order for use in sea ice mass budgets, a diagnostic that is averaged over the ocean grid cell area are needed. The new diagnostic `sidmassgrowthsi` provides this.
-`sidmassgrowthbot` is equal to `LSRCi - sidmassgrowthsi - sidmassgrowthwat` (to float precision) 
-`sidmassmelttop` is equal to `LSNKi - BSNKi - sidmassevapsubl` (to float precision)
-The new diagnostic `sidmassth` is equal to `LSRCi+LSNKi` (to float precision)
 
**Snow on sea ice mass budgets:**
-The new diagnostics for sea ice mass budgets (`d/dt sisnmass = XPRTs + LSNKs + LSRCs`) close the snow on sea ice mass budget at the gridpoint scale with relative errors of roughly 1e-3.
-`sisndmasssi` is equal to the negative of `sidmassgrowthsi`
-`sisndmasssubl+sidmassevapsubl` is equal to `-EVAP-evs`, as expected
-Note that the evaporation and sublimation term is treated as a sink of mass both for snow and sea ice. This is true in all cases I have seen, but in principle this term can be a mass source in the case of frost formation. I think that this choice is ok.
-TO DO: `sisndmassmelt` appears to not be equal to `LSNKs - sisndmasssubl - sisndmasssi`. In particular, the `sisndmassmelt` values have a much larger magnitude than `LSNKs` (e.g. 1e-2 vs 1e-5). Additionally, these values should be negative, but are positive.

**Sea ice concentration budgets:**
-TO DO: I am having difficulty closing the sea ice concentration budget using the new diagnostics. I am using the budget evolution equation `d/dt siconc = XPRTc + LSNKc + LSRCc`. The `XPRTc` term looks to be correct, in that it tracks `siconc` well at the gridpoint scale, but the `LSRCc` term is large and leads to an accumulation of too much ice area when used to reconstruct `siconc`. This appears to not be a simple time unit issue, so I am not sure what is going wrong here.

src/SIS_slow_thermo.F90
```diff
@@ -469,6 +481,39 @@ subroutine slow_thermodynamics(IST, dt_slow, CS, OSS, FIA, XSF, IOF, G, US, IG,
   call adjust_ice_categories(IST%mH_ice, IST%mH_snow, IST%mH_pond, IST%part_size, &
                              IST%TrReg, G, IG, CS%SIS_transport_CSp) !Niki: add ridging?
 
+  do j=jsc,jec ; do k=1,ncat ; do i=isc,iec
+    h2o_change_c(i,j) = h2o_change_c(i,j) + IST%part_size(i,j,k)
+  enddo ; enddo ; enddo
+
+  do j=jsc,jec ; do k=1,ncat ; do i=isc,iec
```
William-gregory: `h2o_change_c` is updated twice on L488-490. This is likely the reason the area budget is not closing @MitchBushuk

MitchBushuk: Thanks Will! I have confirmed that deleting Lines 488-490 fixes the issue. The siconc budgets close to 1e-3 with this change, which is on par with the mass budget closure. @Hallberg-NOAA could you update your branch with this change?

Hallberg-NOAA: I have deleted the duplicated update.  I have also renamed the local variable `h2o_change_c` to `frac_area_change` to better describe what it is used for.

In addition I added a proper commit message to this final commit (based on the description of the PR as a whole above), so that when the commits are squashed together there will be a good description in the commit log.

  At this point, I think that this PR may be ready to be merged in, @MitchBushuk and @William-gregory, although we do need to track down why the diagnosed local budgets are not closing with greater accuracy when the underlying globally integrated budgets are demonstrably closing to better than round-off.

MitchBushuk: Thanks @Hallberg-NOAA. I have a few other bug fixes that Will and I would like to incorporate in, including a fix for the currently broken sisndmassmelt diagnostic. What is the preferred method for this? Should we open a new PR, building off your WG_cmip7_diags_rwh branch?

Hallberg-NOAA:
As long as the solutions themselves are not changing, I think that at this point if you are convinced that this rebased branch correctly captured what was in the original commit, we could merge this into dev/gfdl and then put in a separate PR correcting the calculation of `sisndmassmelt` and the other bug fixes.  This PR is already very large.

MitchBushuk:
I agree that this is a good path forward. We can make the additional fixes in a new PR. I have tested this latest version of the rebased branch, and confirmed that answers are preserved. In my view, this is ready to merge.

MitchBushuk (approved):

Hallberg-NOAA:
This PR has passed pipeline testing at https://gitlab.gfdl.noaa.gov/ogrp/mom6ci/SIS2/-/pipelines/29940 with the expected warnings about new diagnostics.

PR #247, opened 2026-02-02
Add scaling by Idtslow for two mass budget terms

MitchBushuk:
The recently added mass budget diagnostics `sisndmassmelt` and `sidmassgrowthbot` are computed using a difference of other mass budget terms. Some of these terms are missing a scaling by `Idt_slow`. This PR fixes this by adding the correct time scaling. Solutions are bitwise identical, but the diagnostic output will change.

### DIFF
--- a/src/SIS_slow_thermo.F90
+++ b/src/SIS_slow_thermo.F90
@@ -1488,8 +1488,8 @@ subroutine SIS2_thermodynamics(IST, dt_slow, CS, OSS, FIA, IOF, G, US, IG)
       s2i_agg = 0.0
       do k=0,ncat
         s2i_agg = s2i_agg + IST%part_size(i,j,k) * snow_to_ice_s(i,j,k)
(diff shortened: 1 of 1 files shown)


Hallberg-NOAA (approved):
I have examined these changes and I believe them to be correct, and that they fix a bug in the previous code.  These changes will not change solutions, but they will change (and correct) two diagnostics.

  However, there are instances of the sum of three terms in these diagnostics without the use of parentheses to specify the order of arithmetic, as recommended in the MOM6 code style guide at https://github.com/NOAA-GFDL/MOM6/wiki/Code-style-guide#arithmetic-reproducibility.  This should be corrected in a subsequent PR.

Hallberg-NOAA:
This PR has passed pipeline testing at https://gitlab.gfdl.noaa.gov/ogrp/mom6ci/SIS2/-/pipelines/29992.

PR #252, opened 2026-03-11
Fix unpack_ocean_ice_boundary_calved_shelf_bergs 

alex-huth:
In subroutine unpack_ocean_ice_boundary_calved_shelf_bergs, corrected FIA%calving and FIA%calving_hflx so that they are on slow-ice PEs, as called from the coupler (this type of calving comes from the ice shelf on the MOM6 grid).

Also made changes to allow calving in a cell to accumulate from both ice-sheet flux and land model frozen runoff, as there are cases where this should be possible.


Hallberg-NOAA:
This PR has passed pipeline testing at https://gitlab.gfdl.noaa.gov/ogrp/mom6ci/SIS2/-/pipelines/30379.

Hallberg-NOAA (approved):
These changes seem to me to be correct. It should be noted that this routine in not being called yet, although it will be once a pending PR to the FMS coupler (https://github.com/NOAA-GFDL/FMScoupler/pull/147) is resolved.

PR #254, opened 2026-04-16
Bugfix in unit conversion for salt_left_behind

breichl:
- salt_left_behind is used by the brine plume parameterization and passed to an ocean model (MOM6).  It is expected to be in units of kg/m2/s by MOM6.  However, only one of the two terms was being converted to kg here, resulting in mixed units and this amount was roughly 1000x the expected value.  This commit moves the conversion so that the total calculation is converted to kg/m2/s after two terms are subtracted (total salt in water that goes to ice minus fraction of salt that actually goes to ice, leaving the rejected salt).  It will (significantly) change the behavior of the brine plume parameterization in MOM6, but it is approximately compensated by an unrelated bug inside MOM6.


Hallberg-NOAA (approved):
I have carefully reviewed these changes and all of the code that relates to them, and I am convinced that this does correctly fix a bug in the brine plume code.

The `DO_BRINE_PLUME` option is a relatively recent addition to SIS2, and we have been in contact with everyone who we think might have been using it, and we are convinced that this bug can be fixed here without requiring a bug-retention run-time option.

Hallberg-NOAA:
This PR has passed pipeline testing at https://gitlab.gfdl.noaa.gov/ogrp/mom6ci/SIS2/-/pipelines/30784.

Issue #256, opened 2026-04-30
Excessive standard out due to post_data_3d reports


breichl:
When OM5 is coupled to CM4 I am finding I get tens of thousands of lines of standard out that look like this (it seems to average about 100 lines per simulated day):

NOTE from PE    0: callTree:    ---> post_data_3d(ice_model-CN)
NOTE from PE    0: callTree:    <--- post_data_3d(ice_model-CN)

It is probably slowing the model down and is certainly causing a spike in the size of the standard out file.  

@MitchBushuk suggests that CN might be the only 3d output we are storing in this run.  It doesn't happen in standard OM5 runs forced by JRA.  The runs use different diag_tables, but both store CN in ice_month files.  It did not happen in a OM5-CM4 run using the same diag_table but with source code from around September 2025, so it is probably related to changes in SIS2 in the past year.

adcroft:
What is `VERBOSITY` set to?  https://github.com/mom-ocean/MOM6/blob/main/src/framework/MOM_error_handler.F90#L237 suggests it is greater than 5.

breichl:
I don't see a VERBOSITY flag in SIS_parameter_doc.debugging.  I do see a VERBOSE flag, which is False.
The MOM VERBOSITY is set to 2.

MitchBushuk:
There were some changes made to post_data_3d in [PR 229](https://github.com/NOAA-GFDL/SIS2/pull/229/) in Dec 2025, including changes related to call_tree. I suspect this came in there.

breichl:
I think @adcroft identified the issue, which is that verbosity is initialized to 6 in MOM error handler.  It doesn't look like verbosity is ever set by the SIS2 diag manager, so it presumably keeps this initialized value and prints the call tree (even if it is set <6 in MOM and VERBOSE = False in SIS).  The call tree print statements are only in the post_data_3d, they are not in post_data_2d, hence why it only shows in CN.  It looks like the handling of the call tree prints in SIS2 should be improved to fix this issue.  But this doesn't explain why we don't see the issue in JRA forced OM5 runs (could it have something to do with the PE sets that MOM and SIS occupy?).

MitchBushuk:
This issue was resolved with PR #257.

PR #257, opened 2026-05-05
+Add VERBOSITY

Hallberg-NOAA:
Added a runtime `VERBOSITY` parameter to SIS2, with a default value of 2 (for Fatal errors and warnings only), along with a call to `SIS_set_verbosity()` (which is an alias to `MOM_set_verbosity()`) to set the verbosity on sea-ice PEs.  When the sea-ice and ocean are running on the same PEs, the MOM6 verbosity setting takes precedence.  All answers are bitwise identical, but there may be a reduction in the volume of output to standard out in some cases.

  This PR will address the issue noted at https://github.com/NOAA-GFDL/SIS2/issues/256.


MitchBushuk (approved):
This is a useful addition to SIS2, which will allow for better control of the standard out. The changes all look correct to me. Also, I agree that `verbosity=2` is a good choice for the default value. Thanks @Hallberg-NOAA for the fast action on Issue #256!

Hallberg-NOAA:
This PR has passed pipeline testing at https://gitlab.gfdl.noaa.gov/ogrp/mom6ci/SIS2/-/pipelines/30794 with the expected warnings about a new runtime parameter in the SIS_parameter_doc files.

PR #259, opened 2026-05-21
Add ensemble functionality

William-gregory:
This PR adds some functionalities needed for running ensemble experiments with SIS2. 

The changes to `SIS_get_input.F90` mean that the user can run a perturbed sea ice physics ensemble by passing unique `SIS_override` files to `SIS_input_nml` via:

`parameter_filename = 'INPUT/SIS_input','INPUT/SIS_layout','INPUT/SIS_override_%E'`

The changes to `SIS_sum_output.F90` mean that a `seaice.stats` file will be created for each ensemble member, allowing the user to track conservation statistics for each member.

The changes to `SIS_fixed_initialization.F90` mean that a `sea_ice_geometry` file will be created for each ensemble member. This was needed because the simulation crashes during initialization when running a large (30-member) perturbed sea ice physics ensemble from a cold start (`input_filename = 'n'`). It seemed like all the ensemble members were trying to read/write the same `sea_ice_geometry` file. This fix of having one geometry file per member is not the most elegant solution, but at least the model runs.

### DIFF
--- a/src/SIS_get_input.F90
+++ b/src/SIS_get_input.F90
@@ -29,12 +29,14 @@ contains
 
 !> Get_SIS_input reads the SIS namelist entries to see if the run is to be started from
 !! a saved restart file, and get the names of the parameter files, I/O directories.
-subroutine Get_SIS_Input(param_file, dirs, check_params, component)
+subroutine Get_SIS_Input(param_file, dirs, check_params, component, ensemble_num)
   type(param_file_type), optional, intent(out) :: param_file !< A structure to parse for run-time parameters
   type(directories),     optional, intent(out) :: dirs         !< Container for paths and parameter file names.
   logical,               optional, intent(in)  :: check_params !< If present and False will stop error checking for
                                                                !! run-time parameters.
   character(len=*),      optional, intent(in)  :: component    !< An alternate component name, the default is "SIS"
+
+  integer,               optional, intent(in)  :: ensemble_num !< The ensemble id of the current member
 
 !    See if the run is to be started from saved conditions, and get  !
 !  the names of the I/O directories and initialization file.  This   !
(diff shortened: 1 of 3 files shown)


theresa-cordero:
When using this with the Arc_11k simulation and 30 ensemble members using 258 nodes the run was crashing during the writing of the sea_ice_geometry file. 

I think the solution Will suggested would work, but if it was possible to have just the first ensemble member write the sea_ice_geometry file that would be better.

theresa-cordero:
Writing the sea_ice_geometry file can be turned off through on its own, which is the solution I have been using.

William-gregory:
> When using this with the Arc_11k simulation and 30 ensemble members using 258 nodes the run was crashing during the writing of the sea_ice_geometry file.
> 
> I think the solution Will suggested would work, but if it was possible to have just the first ensemble member write the sea_ice_geometry file that would be better.

Yes, good point! I just made this change, so now only ens_01 writes the geom file. Can confirm that this works in a 30-member ensemble job.

src/SIS_get_input.F90
```diff
@@ -79,17 +91,25 @@ subroutine Get_SIS_Input(param_file, dirs, check_params, component)
...
-      endif
+  
+    if (len_trim(trim(parameter_filename(io))) > 0) then
+  
+     if (present(ensemble_num)) then
+            call open_param_file(trim(ensembler(parameter_filename(io),ensemble_num)), param_file, &
+                                 check_params, component=comp, &
+                                 doc_file_dir=output_dir)
+     else
+            call open_param_file(trim(ensembler(parameter_filename(io))), param_file, &
```
Hallberg-NOAA: The second argument to `ensembler()` is optional, so there is no need for the test for whether `ensemble_num` is present here.  The Fortran compiler will handle this for you, so the only change that is needed to the original code is to change line 83 from
```
     call open_param_file(trim(parameter_filename(io)), param_file, &
```
to
```
     call open_param_file(trim(ensembler(parameter_filename(io),ensemble_num)), param_file, &
```

  Also, SIS2 follows the MOM6 code style guide that strongly encourages a 2-point indenting between code levels (see https://github.com/NOAA-GFDL/MOM6/wiki/Code-style-guide#white-space), and not the much larger or missing indents used here.

William-gregory: Thanks Bob, resolving this now

Hallberg-NOAA (approved):
These changes look correct to me.

  Although I am approving them, we will have to wait for the pipeline runners on Gaea to come back up before we can run the tests to verify that this PR is passing the regression testing and then merge them into dev/gfdl.

theresa-cordero (approved):
I tested this code with 30 ensemble members of the Arctic Regional domain and it ran successfully.

William-gregory:
Thanks Theresa and Bob!

Hallberg-NOAA:
This PR has passed SIS2 pipeline testing at https://gitlab.gfdl.noaa.gov/ogrp/mom6ci/SIS2/-/pipelines/30978.

